PR #5, opened 2015-05-01
Prevent errors for Spaces after closing braces, as we want a newline

jdevalk:


Rarst:
Looks good. Originally caused by rule added to WPCS, in line with their (different) if/else handling.

jdevalk:
Let's merg then :)

PR #13, opened 2016-01-07
Made WPCS dependency version fixed

### LINKED ISSUE #12: Prevent new sniffs from creeping in...
Updating the sniffs should be a manual task that should be done once every release cycle, preferably right after release. Let's make sure that every composer install gives the same result in terms of code style...

Rarst:
Fixes #12


Rarst:
Upstream is ahead of 0.6, so subject to update pass later.

PR #17, opened 2016-12-23
Remove exclusion of the ValidVariableName Sniff.

jrfnl:
This sniff was excluded because it complained about the WP core `$cat_ID` variable. This was however fixed in WPCS 0.10.0.

See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/551


jrfnl:
Rebased for merge conflicts

PR #19, opened 2016-12-23
Use a more appropriate ruleset.

### LINKED ISSUE #16: Are the correct rulesets being used ?
WPCS currently has five rulesets:
* `WordPress-Core` - stand-alone, only rules as documented for core
* `WordPress-Docs` - stand-alone, only documentation rules as documented for core
* `WordPress-Extra` - `WordPress-Core` + a number of extra programming/WP best practice kind of rules
* `WordPress-VIP` - `WordPress-Core` + a number of specific rules which plugins and themes have to comply with to be allowed on to the WP VIP platform.
* `WordPress` - `WordPress-Core` + `WordPress-Docs` + `WordPress-Extra` + `WordPress-VIP` (all current rulesets)

Currently Yoastcs includes the `WordPress-VIP` ruleset and the `WordPress-Docs` ruleset.

As far as I know, the Yoast plugins are not deployed on the VIP platform, so do not *have* to comply with the VIP rules - though they do contain some hints/best practices which may be useful to check against.

So having said that, to me, it would seem more logical to check against either `WordPress` (everything) or against `WordPress-Extra` + `WordPress-Docs`.

Opinions ?

jrfnl:
Fixes #16

----

[Edit] The PR has been updated with a review of current extra rule inclusions/exclusions which are no longer needed if the main ruleset changes.

* As the WordPress ruleset includes all WP sniffs, specific sniffs which were not in the VIP ruleset don't need to be included separately anymore.
* As the WordPress ruleset includes the Docs ruleset, no need to include it separately. The related rule exclusions have been moved up to the main ruleset inclusion block (except for the scalar type hint one as that is no longer needed - see #22).
* The GlobalVariables sniff was intended to be added, but commented out for bugs. As it will now be automatically included as the ruleset has changed, it will need to be explicitely excluded. The exclusion can be removed once WPCS 0.11.0 has been released which fixes all known bugs in that sniff. See #20

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -3,7 +3,7 @@
 	<description>Yoast Coding Standards</description>
 
     <!-- ##### WordPress sniffs #####-->
-	<rule ref="WordPress-VIP">
+	<rule ref="WordPress">
 		<exclude name="Generic.Files.LineEndings.InvalidEOLChar" />
 		<exclude name="Squiz.WhiteSpace.SuperfluousWhitespace.EmptyLines" />
 		<!-- Some calls just too quirky and complicated for this. -->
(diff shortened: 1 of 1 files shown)


jrfnl:
Rebased for merge conflicts

PR #27, opened 2017-03-21
:rocket: Minor travis script tweak

jrfnl:
Run the setup in `before_install` instead of `before_script` to report failures in the setup as build `errored` instead of `failed`.

This is important as it determines there was an issue setting up your environment rather than the code you are testing against.


ntwb (approved):
LGTM

PR #32, opened 2017-03-22
Exclude some more VIP restricted functions groups.

jrfnl:
Which would be triggered in WPSEO otherwise.

PR #33, opened 2017-06-12
Add url_to_postid to the restrictedfunctions

andizer:


jrfnl:
@andizer Could you add this change to `JRF/add-some-more-VIP-exclusions` branch or pull against that instead ? Otherwise this will conflict with #32.

Also: why is it being excluded ?

andizer:
Oh yes, I can add this to that pull.

We need to add this because we are implementing a function that gets an post id for given url but I get an error. By adding this to the excluded restrictedfunction it will be solved.

jrfnl:
> we are implementing a function that gets an post id for given url but I get an error. By adding this to the excluded restrictedfunction it will be solved.

@andizer Getting an error for which plugin ? Last time I looked WPSEO for instance was still on YoastCS 0.3, so wouldn't (yet) benefit until the YoastCS version would be updated.
(don't mean to be a pain, but just making sure the change is properly documented here)

andizer:
I am working on code for WPSEO that needs the `url_to_postid` function, but running the code locally results in an error. 
> url_to_postid is prohibited, please use wpcom_vip_url_to_postid() instead.

That's why I've created this pull.

jrfnl:
FYI: PR https://github.com/Yoast/wordpress-seo/pull/6875 is open to update WPSEO to start using YoastCS 0.4

moorscode:
Added the entry to the https://github.com/Yoast/yoastcs/pull/32 pull,

PR #42, opened 2017-08-02
Update YoastCS to use WPCS 0.11.0

### LINKED ISSUE #20: Re-enable the WordPress.Variables.GlobalVariables sniff
The `WordPress.Variables.GlobalVariables` sniff was originally part of the intended ruleset, but turned off because of its buggyness as reported in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/300.
See: https://github.com/Yoast/yoastcs/blob/master/Yoast/ruleset.xml#L48

All the currently known bugs are expected to be fixed in the near future once https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/736 is merged.

It is expected that the PR will be part of the next WPCS release 0.11.0.

So I'd like to suggest re-enabling the sniff once WPCS 0.11.0 has been released and the minimum WPCS requirement in the Yoast CS `composer.json` has been  upped to 0.11.0.

jrfnl:
* Use WPCS 0.11.0
* Don't fix PHPCS at a specific version.
* Remove commented out exclusion of the WordPress.Variables.GlobalVariables sniff.
   The bugs the comment refers to have been fixed in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/736 which is included in WPCS 0.11.0.
   Additionally, unit tests may need to overload global variables. To that end the known unit test case classes for Yoast plugins have been whitelisted.
* Exclude "Closing comment for long conditions" sniff as it will very quickly become annoying and will be removed in WPCS 0.12.0 anyway.
* Exclude the "associative arrays (i.e. arrays with defined keys) must always be multi-line" sniff.
* Make the filename check less strict for classes.
* Minor adjustment to the group exclusions for WPCS 0.11.0
    In WPCS 0.11.0 the sniffs which check for usage of certain functions have been re-arranged, including some changes in error level (warning vs error).
    This means that some exclusions are no longer needed for the builds to pass and others needed to be added.
* Remove the customization for the string concat sniff as this is now included in WPCS itself.

For more information about the properties used:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#excluding-a-group-of-checks

For further details, see the individual commits.

Fixes #20

### TO DO before merge:

- [x] Run `composer update` and add refreshed `composer.lock` file

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -27,12 +27,6 @@
 		<!-- Turned off because of known & reported bugs in the Sniffs, should be turned on once the bugs are fixed. -->
 		<exclude name="WordPress.VIP.ValidatedSanitizedInput"/>
 
-		<!-- Turned off because of known & reported bugs in the Sniffs, should be turned on once the bugs are fixed. -->
-		<!-- Exclusion can be removed once the WPCS 0.11.0 has been released - in which the bugs have been fixed -
-		     and the minimum WPCS version required by Yoast CS has been upped to 0.11.0. -->
-		<!-- WPCS #300 -->
-		<exclude name="WordPress.Variables.GlobalVariables"/>
-
 		<!-- Catches way too many things, like vars and file headers. -->
 		<exclude name="Generic.Commenting.DocComment.MissingShort"/>
 	</rule>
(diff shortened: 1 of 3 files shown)


jrfnl:
Closing in favour of an upcoming more comprehensive PR which will update to WPCS 0.14.0 in one go as discussed with @moorscode .

Issue #15, opened 2016-09-19
Re-enable PHPCompatibility Sniffs


jrfnl:
I noticed that the [PHPCompatibility](https://github.com/wimg/PHPCompatibility) ruleset has been disabled as it - at the time - did not (yet) support PHPCS 2.x. This has been fixed for over a year now and since then a fairly complete suite of tests has been added to the PHPCompatibility library for PHP 7.

I'd like to suggest re-enabling it.

To that end, I've set up the basics for this in a feature branch, but this will need testing, preferably by someone a lot better versed in Composer than me ;-) _hint hint nudge nudge_

See: https://github.com/Yoast/yoastcs/tree/feature/re-enable-phpcompatibility

Rarst:
As per https://github.com/wimg/PHPCompatibility/pull/234 it's sort of broken for Composer use at the moment.

I really don't like that plugin approach to move folder after install and it hadn't been _necessary_ with support for relative paths in PHP CS config.

jrfnl:
> As per wimg/PHPCompatibility#234 it's sort of broken for Composer use at the moment.

There are a few issues open regarding that, but for most Composer installs - this issue does not apply. From all I've read about the issue, it's mostly a problem on Windows, so doesn't apply to travis runs.

> I really don't like that plugin approach to move folder after install and it hadn't been necessary with support for relative paths in PHP CS config.

Not sure what you mean exactly. Could you elaborate a bit ?

Rarst:
It's not a problem on Windows, it's quite clearly a problem with current plugin setup. Its `autoload` directive in `composer.json` is effectively non-functional.

The way it _does_ work is using that additional plugin to physically move the standard inside PHP CS directory. I would say that is highly not in line with normal Composer mechanics and isn't necessary with current PHP CS versions. As you can see in our `composer.json` we just run configuration command which tells PHP CS where the standards are located in vendor.

Rarst:
TL;DR its Composer setup is part broken, part unorthodox and I am not too keen to drag that into YoastCS and deal with it. :)

jrfnl:
I see what you mean. That is actually a discussion I'm trying to resolve in the PHPCompatibility repo already. I've proposed a potential solution which would be more in line for Composer (moving the Sniffs one directory down), but as it would break any PEAR installs, it is not considered an acceptable solution ATM.

If you have any input on how to solve this better - it would be very welcome.

[...]

PR #43, opened 2017-08-15
Re-enable PHPCompatibility sniffs.

jrfnl:
Rationale:
* PHPCompatibility now runs on PHPCS 2.x without problems (and has been doing so for quite a long time).
* PHPCompatibility issues with Composer have been solved with the release of v 8.0.0.
* PHPCompatibility 8.0.0 also brings cross-version compatibility with PHPCS 3.x so won't hold YoastCS back when it wants to upgrade.

Additionally:
* Set `testVersion` to receive any messages relevant for PHP 5.2+.
* The `DealerDirect` composer plugin which has previously been added as a suggestion for projects including YoastCS should sort out the `installed_paths` config setting for PHPCS instead of doing this via a script in projects using YoastCS.

Fixes issue #15

- [x] Run `composer install` and refresh `composer.lock` file

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -57,10 +57,8 @@
 	</rule>
 
 	<!-- ##### Sniffs for PHP cross-version compatibility ##### -->
-	<!-- Currently disabled because it does not play nice with Composer.
-	     This is being worked on in https://github.com/wimg/PHPCompatibility/issues/102 -->
-	<!--<config name="testVersion" value="5.2-99.0"/>-->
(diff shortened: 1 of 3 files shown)


moorscode:
I'm trying to test this, but `phpcs` is telling me:
```
Fatal error: Uncaught exception 'PHP_CodeSniffer_Exception' with message 'Referenced sniff "PHPCompatibility" does not exist' in wordpress-seo/vendor/squizlabs/php_codesniffer/CodeSniffer.php:1167
```

Is this depending on https://github.com/Yoast/yoastcs/pull/42 or another commit perhaps?

PR #50, opened 2017-11-03
Update YoastCS to PHPCS 3.x/WPCS 0.14.0

### LINKED ISSUE #36: Update for PHPCS 3.x compatibility
Once WPCS 0.13.0 has been released (expected soon), WPCS will be compatible with PHPCS 3.x.

To upgrade YoastCS to start using PHPCS 3.x, the custom sniff will need to be adjusted.

jrfnl:
* Up the required version of PHPCS and WPCS
* Adjust the Travis script to allow for changes in PHPCS 3.x
* Adjust the YoastCS native script for compliance with PHPCS 3.x

See the individual commits for more details.

Fixes #36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,57 +18,36 @@ php:
     - nightly
 
 env:
-  # Test against the highest supported PHPCS version (2.x dev).
-  # This can be changed back to "master" once PHPCS 3.x is supported. See issue #36
-  - PHPCS_BRANCH=2.9
-  # Test against the lowest supported PHPCS 2.x version.
-  - PHPCS_BRANCH=2.8.1
+  # Test against the highest supported PHPCS version.
+  - PHPCS_BRANCH=master
+  # Test against the lowest supported PHPCS version.
+  - PHPCS_BRANCH=3.1.1
 
 matrix:
   fast_finish: true
   include:
-    # PHP 5.2/5.3 need precise.
-    - php: 5.2
-      dist: precise
-      env: PHPCS_BRANCH=2.9
-    - php: 5.2
-      dist: precise
-      env: PHPCS_BRANCH=2.8.1
-    - php: 5.3
-      dist: precise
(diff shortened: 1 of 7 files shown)


moorscode:
I'm a bit concerned that if we require PHP 5.4 here, using this package in Yoast SEO will cause problems when doing a `composer install` during the travis build on PHP < 5.4.

Do we need to remove that dependency when running travis for PHP < 5.4 before doing an install?

Note: haven't tested this.

jrfnl:
> I'm a bit concerned that if we require PHP 5.4 here, using this package in Yoast SEO will cause problems when doing a composer install during the travis build on PHP < 5.4.

Good question. Bit difficult to test as branches which aren't linked to a PR can't be build on Travis with the settings the repos currently have.

> Do we need to remove that dependency when running travis for PHP < 5.4 before doing an install?

I've been looking at the Travis script and the Composer dependencies for Free and Premium.
Your suggestion would definitely be an option, but maybe simpler would be to do something along the lines of:
```
- if [[ "$PHPCS" != "1" && "$COVERAGE" != "1" ]]; then composer install --no-dev --no-interaction; fi
- if [[ "$PHPCS" == "1" ||  "$COVERAGE" == "1"  ]]; then composer install --no-interaction; fi
- if [[ "$PHPCS" == "1" ]]; then composer config-yoastcs; fi
```

AFAICS, the `codeclimate/php-test-reporter` should only be needed for the build testing code coverage - which is currently done in the `7.0` build - same build as the PHPCS check.
AFAICS, the `yoast/php-development-environment` dev dependency is not actually needed for the Travis build, though I might be missing something.

What do you think ?

moorscode:
Implementing a conditional to apply the `--no-dev` seems like a nice approach, this should even improve the other builds in terms of speed.

jrfnl:
In that case, this PR should be good to merge.

The Travis script adjustments should be made in the plugin specific Travis files if/when each plugin upgrades to YoastCS 0.5.

PR #52, opened 2017-11-30
YoastCS: check YoastCS native files for codestyle compliance

jrfnl:
Includes addition of a fully documented `.phpcs.xml.dist` file specifically for YoastCS itself and fixes to the IfElseDeclaration sniff to comply.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -1,0 +1,41 @@
+<?xml version="1.0"?>
+<ruleset name="Yoast Coding Standard">
+	<description>The Coding standard for the Yoast Coding Standard itself.</description>
+
+	<!-- Scan all files. -->
+	<file>.</file>
+
+	<!-- Exclude Composer vendor directory. -->
+	<exclude-pattern>*/vendor/*</exclude-pattern>
+
+	<!-- Only check PHP files. -->
+	<arg name="extensions" value="php"/>
+
+	<!-- Show progress, show the error codes for each message (source). -->
+	<arg value="sp"/>
+
+	<!-- YoastCS supports PHP 5.4 or higher. -->
+	<config name="testVersion" value="5.4-"/>
+
+	<rule ref="Yoast">
+
+		<!-- Conflicts with PHPCS autoloading of sniffs. -->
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -42,12 +42,24 @@ before_install:
     - export XMLLINT_INDENT="	"
     - export PHPCS_DIR=/tmp/phpcs
     - export PHPCS_BIN=$PHPCS_DIR/bin/phpcs
+    - export WPCS_DIR=/tmp/wpcs
+    - export PHPCOMPAT_DIR=/tmp/phpcompatibility
     - mkdir -p $PHPCS_DIR && git clone --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git -b $PHPCS_BRANCH $PHPCS_DIR
     - $PHPCS_BIN --config-set installed_paths $(pwd)
```
moorscode, line 48: Is this line still needed? It seems like the configuration is set twice when `$SNIFF` is configured.
When not sniffing, this does not need to be cloned or configured right?

.phpcs.xml.dist
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0"?>
+<ruleset name="Yoast Coding Standard">
+	<description>The Coding standard for the Yoast Coding Standard itself.</description>
+
+	<!-- Scan all files. -->
+    <file>.</file>
```
moorscode: Indentation seems off.

moorscode (changes requested):
Possible small optimization change for cloning/checking out PHPCS.

.travis.yml
jrfnl: When not sniffing, the unit tests are run using the PHPCS unit test framework. For that to *only* test the YoastCS native sniffs, the YoastCS standard needs to be registered, so yes, this is still needed, though it could possibly be combined with the version of it below with an if/then/else syntax. You know how to do those in Travis scripts ?

.phpcs.xml.dist
jrfnl: Fixed up.

PR #54, opened 2017-11-30
PHPCS Ruleset updates

### LINKED ISSUE #20: Re-enable the WordPress.Variables.GlobalVariables sniff

### LINKED ISSUE #35: Update needed for WPCS 0.12.0
WPCS 0.12.0 was released last week.

A number of new sniffs were introduced and some more custom properties which YoastCS could benefit from.

jrfnl:
As WPCS `0.14.0` is now the target version, some changes to the ruleset can be made to:
* Remove rule duplication - for sniff(s) which are now included upstream
* Remove outdated exclusions - for sniff which used to be buggy, which has been fixed
* Add configuration for a new WPCS sniff
* Exclude warnings for two specific situations.

See the individual commits for more details.

Fixes #20
Fixes #35

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -17,6 +17,17 @@
 
 	</rule>
 
+	<rule ref="WordPress.Arrays.MultipleStatementAlignment">
+		<properties>
+			<!-- No need to adjust alignment of large arrays when the item with the largest key is removed. -->
+			<property name="exact" value="false"/>
+			<!-- Don't align multi-line items if ALL items in the array are multi-line. -->
+			<property name="alignMultilineItems" value="!=100"/>
+			<!-- Array Assignment operator should always be on the same line as the array key. -->
+			<property name="ignoreNewlines" value="false"/>
+		</properties>
+	</rule>
+
 	<!-- ##### Sniffs for PHP cross-version compatibility ##### -->
 	<config name="testVersion" value="5.2-99.0"/>
 	<rule ref="PHPCompatibility"/>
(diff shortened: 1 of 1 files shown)


Yoast/ruleset.xml
```diff
@@ -15,11 +15,22 @@
...
 
-		<!-- Turned off because of known & reported bugs in the Sniffs, should be turned on once the bugs are fixed. -->
-		<!-- Exclusion can be removed once the WPCS 0.11.0 has been released - in which the bugs have been fixed -
-		     and the minimum WPCS version required by Yoast CS has been upped to 0.11.0. -->
-		<!-- WPCS #300 -->
-		<exclude name="WordPress.Variables.GlobalVariables"/>
+		<!-- A while loop is the only valid control structure where an assignment can be justified. -->
+		<exclude name="WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition"/>
+
+		<!-- If a conscious choice has been made for a non-strict comparison, that's ok. -->
```
moorscode: Perhaps good to mention the specific situation for this comment.
"When strict has been explicitly set to `false` in an `in_array`-call, that should be allowed."

moorscode (changes requested):
More concrete/explicit explanation about the specific sniff.

Yoast/ruleset.xml
jrfnl: Done.

PR #56, opened 2017-12-21
Add new YoastCS native FileName sniff

### LINKED ISSUE #48: YoastCS native filename sniff
As discussed with @moorscode and by @moorscode discussed with @omarreiss and @atimmer, I'm working on a YoastCS native filename sniff.

There are a couple of things I need to get a decision on before the sniff can be finalized, so Jip, Omar, Anton, please discuss and let me know the outcome.

Currently implemented rules:
* (WP) Filenames should be lowercase and words should be separated by dashes (not underscores).
* All class files should only contain one class (enforced by another sniff in WPCS) and the file name should reflect the class name without the plugin specific prefix.
* All interface and trait files should only contain one interface/trait (enforced by another sniff in WPCS) and the file name should reflect the interface/trait name without the plugin specific prefix and with an `-interface` or `-trait` suffix.
* Files which don't contain an object structure, but do contain one or more function declarations should have a `-functions` suffix.

Currently implemented exception:
* It will be possible to pass one or more file names to the sniff to exclude.
    This has been done to allow for "main" plugin files which may contain functions, but don't have the `-functions` suffix. Renaming the file would lead to deactivation of the plugin on upgrade which is probably not a good idea.

Open questions:
* For the file names to be excluded, should this just be a file name or should this be a path from the root of a plugin and strictly checked against the path ? If it's just a file name, you run the risk that another file in another directory would use the same name and be excluded from this sniff as well.
* Should this sniff be set up to be able to also check the filenames of CSS and JS files ?
* Should the "functions" rule be applied when a file **_only_** contains functions (no side-effects) or also when the file contains functions as well as miscellaneous code ?
    Think: WPSEO Free `admin/pages/metas.php`.
* What about interfaces/classes/traits which have the word `Interface/Class/Trait` in the object name ?
    Think: WPSEO Free `WPSEO_Sitemap_Cache_Data_Interface`
* Should there be a rule about abstract classes ? Should they have an `abstract` prefix/suffix or should this just be the same as the class name ?
* Keeping in mind the intention to move to a PSR based namespace folder structure - should a potential overlap with the directory name a class lives in be removed ?
    Think: WPSEO Free class `WPSEO_Metabox_Tab_Section` which currently lives in file `admin/metabox/class-metabox-tab-section.php`:
    Should this become `metabox-tab-section.php` as per the above rules or should this become `tab-section.php` ?
   And if the latter: what about classes which are named the same as their directory path ? Think `WPSEO_Tracking` in `admin/tracking`.
* Test class files, now all start with `test-class`. For PHPUnit, having the `test-` prefix (or a `-test` suffix) is needed.
    - Should the `class-` bit stay or go ?
    - Should the filename match the class name more closely ? The order is reversed at the moment.
        Class names for test classes generally follow the following pattern: `Prefix_Classname_Test` with the `Prefix` and the `Test` suffix being fairly consistently applied and `Classname` reflecting the name of the class under test.
        File names for test classes generally follow the following pattern at this time: `test-class-classname.php`.
        Using the "normal" rules for classes, the file name should/would become: `classname-test.php`.
        Example:
              - Current file name: `test-class-redirect-table.php`
              - Class name: `WPSEO_Redirect_Table_Test`
              - Expected file name using "normal" rules: `redirect-table-test.php`
    - In other word; how would you like me to handle test files ?
    - And what to do about test classes which are inconsistently named ? Think: `Test_Yoast_Notification`.
* And what about test Mock/Double classes ? Should they follow the file name rules for "normal" classes ?

Reading back over my own questions, I'm thinking that maybe there should be a YoastCS `ClassName` sniff as well to prevent some of the classname inconsistencies I've listed about.

Maybe something along the lines of:
* Don't use `interface`, `class`, `trait` or `abstract` in the object name.
* OO Structures in the plugin should follow the following pattern: `Prefix_[Optional:Subdir_]Classname` where a couple of "basepaths" can be taken into account, such as `admin` or `premium` which don't need to be in the name.
* Test OO Structures do not need a prefix (as they won't be shipped with the plugin) and should follow the following pattern: `[Optional:Subdir_]Classname_Test`.

What do you think ?

It won't be much work to write this sniff, providing the specs are clear.

### Attachments

Current file rename proposal based on the above:
[wpseo-file-rename-proposal.xlsx](https://github.com/Yoast/yoastcs/files/1400732/wpseo-file-rename-proposal.xlsx)

Notes:
* Take note of the two tabs - one for the WPSEO Free plugin, one for the Premium plugin.
* For this list, I've taken two class prefixes into account: `WPSEO` and `Yoast`.
* For the `Free` plugin, the list will not be entirely correct yet as there are still 26 files (2 in the plugin, 24 in the tests) in Free which contain two or more classes and need to be split.
* For the `Premium` plugin, the above mentioned split has already been done.

Raw data in CSV format in case you want to have a play with it (`.csv` files can't be uploaded here, so renamed to `txt`):
[wpseo-free-file-rename.txt](https://github.com/Yoast/yoastcs/files/1400735/wpseo-free-file-rename.txt)
[premium-file-rename.txt](https://github.com/Yoast/yoastcs/files/1400734/premium-file-rename.txt)

jrfnl:
Ensures files comply with the Yoast file name rules.

Rules:
- (WP) Filenames should be lowercase and words should be separated by dashes (not underscores).
- All class files should only contain one class (enforced by another sniff) and the file name should reflect the class name without the plugin specific prefix.
- All interface and trait files should only contain one interface/trait (enforced by another sniff) and the file name should reflect the interface/trait name without the plugin specific prefix and with an "-interface" or "-trait" suffix.
- Files which don't contain an object structure, but do contain function declarations should have a "-functions" suffix.

The sniff allows to `exclude` select files from the last three rules by setting a property in a custom ruleset.
Similarly, the plugin specific prefixes to take into account can also be passed in through a custom ruleset.

Includes extensive unit  tests.

Fixes #48


moorscode:
I have loaded this branch in WPSEO Premium and ran the sniff.
Getting unexpected results:

>  1 | ERROR | Class file names should be based on the class name without the plugin prefix. Expected wpseo-validation-warning.php, but found validation-warning.php. (Yoast.Files.FileName.InvalidClassFileName)

The message contradicts itself, which is strange.

Should I do any additional configuration in the CS configuration in Premium to make this work?

I think I figured out that the sniff does not know that the prefix should be `wpseo` or `yoast`.

jrfnl:
@moorscode Correct, the sniff needs some minimal input to work correctly. For premium you'll need to add the following to get the correct results (will be included in the PR to update the Premium ruleset + YoastCS version):

```xml
	<rule ref="Yoast.Files.FileName">
		<properties>
			<!-- Don't trigger on the main file as renaming it would deactivate the plugin. -->
			<property name="exclude" type="array" value="wp-seo-premium.php"/>
			<!-- Remove the following prefixes from the names of object structures. -->
			<property name="prefixes" type="array" value="wpseo,yoast"/>
		</properties>
	</rule>
```

moorscode:
Applied and verified that it gives the expected results.

PR #63, opened 2018-01-12
Travis: speed up build times by disabling Xdebug

jrfnl:
As suggested by John Blackbourn in https://johnblackbourn.com/reducing-travis-ci-build-times-for-wordpress-projects/

Looks like for this simple build script it doesn't make much difference, but it doesn't hurt either,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,10 @@ matrix:
     - php: nightly
 
 before_install:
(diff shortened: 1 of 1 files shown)


jrfnl:
As this PR hasn't been looked at yet, I've amended the original commit to further stabilize the removal command as suggested here: https://twitter.com/kelunik/status/954242454676475904 .

PR #66, opened 2018-01-14
.gitattributes: sync with files currently in the repo

jrfnl:
The `export-ignore` directives in the `.gitattributes` file determine which files are excluded when the repository is packaged for distribution, like on tagging a release.

The file exclusion list was out of date.

Also added the `text=auto` directive to normalize line-endings.


jcomack:
CR & acceptance done 👍

PR #67, opened 2018-01-22
IfElseDeclaration: Various improvements

jrfnl:
This PR addresses a number of issues with the `IfElseDeclaration` sniff.

I initially started reviewing the sniff to make sure it would deal correctly with the new PHPCS 3.2.0+ annotations.
It was cute to see what I did when I really didn't know much about writing sniffs yet :heart_eyes_cat: (This was the first sniff I ever wrote.)

Based on the review, I've made extensive changes to the sniff. Each change is isolated in an individual commit for easier review and traceable history.

Summary of the changes:
* Bug fix: the sniff would presume the WP code style. If code was not using the WP code style, the sniff would not correctly detect issues, effectively leading to "issues hiding behind other issues".
    This fix also sorts out the false positives which would be thrown for the PHPCS annotations.
* Bug fix: the same - issues hiding behind each other - was true for the various error messages this sniff could throw.
* Enhancement: minor improvement to the error message.
* Minor efficiency fixes.
* Tests: The unit tests didn't cover all possible error messages.
* Tests: The unit tests presumed the WP code style.
* Tests: New test to cover the new PHPCS annotations are handled correctly.

### DIFF
--- a/Yoast/Sniffs/ControlStructures/IfElseDeclarationSniff.php
+++ b/Yoast/Sniffs/ControlStructures/IfElseDeclarationSniff.php
@@ -68,7 +68,7 @@ class IfElseDeclarationSniff implements Sniff {
 			return;
 		}
 
-		$previous = $phpcsFile->findPrevious( T_CLOSE_CURLY_BRACKET, $stackPtr, null, false );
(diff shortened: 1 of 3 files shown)


moorscode:
Looks good!

PR #72, opened 2018-03-15
Security: add Roave/SecurityAdvisories dependency

jrfnl:
This dependency will prevent packages with known security issues from being installed through Composer.

The package will always have to be at `dev-master` to make sure that the latest security information available will be used.

Refs:
* https://github.com/Roave/SecurityAdvisories
* https://www.bleepingcomputer.com/news/security/php-community-takes-steps-to-stop-installation-of-libraries-with-unpatched-bugs/
* https://websec.io/2018/03/10/Package-Protection-Roave-SecurityAdvisories.html

-----

While not directly related, I noticed that both `PHP_CodeSniffer` as well as WPCS were locked at outdated versions. As this repo is directly related to those two dependencies, I've updated them both to the latest versions:
* PHP_CodeSniffer 3.2.2 -> 3.2.3
* WPCS 0.14.0 -> 0.14.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,5 +36,8 @@
 		],
 		"post-install-cmd": "composer config-set",
 		"post-update-cmd": "composer config-set"
+	},
+	"require-dev": {
+		"roave/security-advisories": "dev-master"
 	}
 }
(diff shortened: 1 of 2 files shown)


jrfnl:
Regarding the build failure - see my remark about that in #71. Same applies here.

PR #76, opened 2018-03-30
Introduce Required Parameter Sniff

### LINKED ISSUE #74: Create a sniff to ensure `update_option` sets `autoload` explicitly
As the `autoload` setting has been added in a later state, the argument is not required.
Though this is a setting that should be set explicitly, as it impacts performance and memory.

moorscode:
Initially only checking `update_option` for the `autoload` parameter to be set.
This is an optional parameter, but we want to make sure it is set with intent.

As this class is based upon the `AbstractFunctionParameterSniff`-class from WPCS, some class-aliases need to be loaded to make it work.
These are being provided in the autoload-bootstrap.php file.

To make sure the tests also know where to find them, a test bootstrap file had to be added as well.

@jrfnl do I need to give props for the inspirational files and extended class and if so, how?

Fixes #74

### DIFF
--- a/Tests/bootstrap.php
+++ b/Tests/bootstrap.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * Bootstrap file for running the tests.
+ *
+ * Load the PHPCS autoloader and the WPCS PHPCS cross-version helpers.
+ *
+ * Code has been based on the phpcs3-bootstrap.php used in the WordPress Coding Standards project
+ *
+ * @link: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ *
+ * @package Yoast\YoastCS
+ * @link    https://github.com/Yoast/yoastcs
+ * @license https://opensource.org/licenses/MIT MIT
+ * @since   1.1.0
+ */
+
+if ( ! defined( 'PHP_CODESNIFFER_IN_TESTS' ) ) {
+	define( 'PHP_CODESNIFFER_IN_TESTS', true );
+}
+
+/**
+ * Builds the directory path for the given parts.
+ *
+ * @param array $parts Parts to convert to a path.
+ *
+ * @return string The path usable for the current system.
+ */
+function yoastcs_test_bootstrap_get_path( $parts ) {
+	return implode( DIRECTORY_SEPARATOR, $parts );
+}
+
+$ds              = DIRECTORY_SEPARATOR;
+$baseComposerDir = dirname( __DIR__ );
+
+// Get the PHPCS dir from an environment variable.
+$phpcsDir         = getenv( 'PHPCS_DIR' );
+$phpcsComposerDir = yoastcs_test_bootstrap_get_path(
+	array( $baseComposerDir, 'vendor', 'squizlabs', 'php_codesniffer' )
+);
+
+// This may be a Composer install.
+if ( false === $phpcsDir && is_dir( $phpcsComposerDir ) ) {
+	$phpcsDir = $phpcsComposerDir;
+}
+elseif ( false !== $phpcsDir ) {
+	$phpcsDir = realpath( $phpcsDir );
+}
+
+// Try and load the PHPCS autoloader.
+if ( false !== $phpcsDir && file_exists( yoastcs_test_bootstrap_get_path( array( $phpcsDir, 'autoload.php' ) ) ) ) {
+	require_once yoastcs_test_bootstrap_get_path( array( $phpcsDir, 'autoload.php' ) );
+
+	/*
+	 * As of PHPCS 3.1, PHPCS support PHPUnit 6.x, but needs a bootstrap, so
+	 * load it if it's available.
+	 */
+	if ( file_exists( yoastcs_test_bootstrap_get_path( array( $phpcsDir, 'tests', 'bootstrap.php' ) ) ) ) {
+		require_once yoastcs_test_bootstrap_get_path( array( $phpcsDir, 'tests', 'bootstrap.php' ) );
+	}
+}
+else {
+	echo 'Uh oh... can\'t find PHPCS.
+
+If you use Composer, please run `composer install`.
+Otherwise, make sure you set a `PHPCS_DIR` environment variable in your phpunit.xml file
+pointing to the PHPCS directory.
+';
+
+	die( 1 );
+}
+
+// Get the WPCS dir from an environment variable.
+$wpcsDir         = getenv( 'WPCS_DIR' );
+$wpcsComposerDir = yoastcs_test_bootstrap_get_path( array( $baseComposerDir, 'vendor', 'wp-coding-standards', 'wpcs' ) );
+
+// This may be a Composer install.
+if ( false === $wpcsDir && is_dir( $wpcsComposerDir ) ) {
+	$wpcsDir = $wpcsComposerDir;
+}
+elseif ( false !== $wpcsDir ) {
+	$wpcsDir = realpath( $wpcsDir );
+}
+
+// Try and load the WPCS class aliases file.
+$wpcsAliasFilename = yoastcs_test_bootstrap_get_path( array( $wpcsDir, 'WordPress', 'PHPCSAliases.php' ) );
+if ( file_exists( $wpcsAliasFilename ) ) {
+	require_once $wpcsAliasFilename;
+}
+else {
+	echo 'Uh oh... can\'t find WPCS.
+
+If you use Composer, please run `composer install`.
+Otherwise, make sure you set a `WPCS_DIR` environment variable in your phpunit.xml file
+pointing to the WPCS directory.
+';
+
+	die( 1 );
+}
+
+unset( $ds, $baseComposerDir, $phpcsDir, $phpcsComposerDir, $wpcsDir, $wpcsComposerDir, $wpcsAliasFilename );
--- a/Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
+++ b/Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
@@ -73,7 +73,7 @@ class RequiredOptionalParametersSniff extends AbstractFunctionParameterSniff {
 	 */
 	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
 		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
-			// Check that number of parameters defined is not less than the position to check.
+			// Check that a valid parameter value is being passed.
 			if ( $this->has_valid_parameter( $parameters, $position, $parameter_args ) ) {
 				continue;
 			}
--- a/Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
+++ b/Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
@@ -12,8 +12,8 @@ namespace YoastCS\Yoast\Sniffs\WP;
 use WordPress\AbstractFunctionParameterSniff;
 
 /**
- * Sniff that ensures the optional parameter for update_option and set_option
- * are defined because we want developers to be conscious about what is needed.
+ * Sniff that ensures the optional parameter for update_option() and add_option()
+ * are passed because we want developers to be conscious about what is needed.
  *
  * @package Yoast\YoastCS
  * @author  Jip Moors
@@ -38,7 +38,7 @@ class RequiredOptionalParametersSniff extends AbstractFunctionParameterSniff {
 	 *            (int) Target parameter position, 1-based. => array(
 	 *                'name'       => (string)  The name of the argument that should be set.
 	 *                'enforce'    => (boolean) If this should trigger an error or warning.
-	 *                'allow_null' => (boolean) If `null` is seen as setting the optional value, defaults to False.
+	 *                'allow_null' => (boolean) Whether `null` is seen as setting the optional value. Defaults to false.
 	 *            )
 	 *         )
 	 *    );
--- a/Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
+++ b/Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
@@ -1,0 +1,146 @@
+<?php
+/**
+ * YoastCS\Yoast\Sniffs\WP\RequiredOptionalParameters.
+ *
+ * @package Yoast\YoastCS
+ * @author  Jip Moors
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace YoastCS\Yoast\Sniffs\WP;
+
+use WordPress\AbstractFunctionParameterSniff;
+
+/**
+ * Sniff that ensures the optional parameter for update_option and set_option
+ * are defined because we want developers to be conscious about what is needed.
+ *
+ * @package Yoast\YoastCS
+ * @author  Jip Moors
+ *
+ * @since   1.1.0
+ */
+class RequiredOptionalParametersSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'wp_required_parameters';
+
+	/**
+	 * Functions this sniff is looking for.
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *    $target_functions = array(
+	 *        (string) Function name. => array(
+	 *            (int) Target parameter position, 1-based. => array(
+	 *                'name'       => (string)  The name of the argument that should be set.
+	 *                'enforce'    => (boolean) If this should trigger an error or warning.
+	 *                'allow_null' => (boolean) If `null` is seen as setting the optional value, defaults to False.
+	 *            )
+	 *         )
+	 *    );
+	 */
+	protected $target_functions = array(
+		'update_option' => array(
+			3 => array(
+				'name'       => 'autoload',
+				'enforce'    => false,
+				'allow_null' => false,
+			),
+		),
+		'add_option'    => array(
+			4 => array(
+				'name'       => 'autoload',
+				'enforce'    => true,
+				'allow_null' => false,
+			),
+		),
+	);
+
+	/**
+	 * Process the parameters of a matched function.
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+			// Check that number of parameters defined is not less than the position to check.
+			if ( $this->has_valid_parameter( $parameters, $position, $parameter_args ) ) {
+				continue;
+			}
+
+			$message = '%s() is %s to be called with the "%s" argument at position %d%s.';
+
+			$error_type = ( $parameter_args['enforce'] === true ) ? 'Required' : 'Optional';
+			$error_null = $this->is_null_allowed( $parameter_args ) ? '' : 'NullDisallowed';
+
+			$error_string = $matched_content . $error_type . 'Param' . ucfirst( $parameter_args['name'] ) . 'Missing' . $error_null;
+
+			$code = $this->string_to_errorcode( $error_string );
+
+			$data = array(
+				$matched_content,
+				( $parameter_args['enforce'] === true ) ? 'required' : 'expected',
+				$parameter_args['name'],
+				$position,
+				( $this->is_null_allowed( $parameter_args ) ) ? '' : ', which cannot be "null"',
+			);
+
+			$is_error = $parameter_args['enforce'];
+			if ( ! $is_error && ! $parameter_args['allow_null'] ) {
+				$is_error = ( isset( $parameters[ $position ] ) && $parameters[ $position ]['raw'] === 'null' );
+			}
+
+			$this->addMessage( $message, $stackPtr, $is_error, $code, $data );
+		}
+	}
+
+	/**
+	 * Determines if a valid parameter has been provided.
+	 *
+	 * @param array $parameters     The parameters used to call the function.
+	 * @param int   $position       The position of the parameter to check.
+	 * @param array $parameter_args The configuration of the function check.
+	 *
+	 * @return bool True if a valid parameter has been given, False otherwise.
+	 */
+	protected function has_valid_parameter( $parameters, $position, $parameter_args ) {
+		if ( ! isset( $parameters[ $position ] ) ) {
+			return false;
+		}
+
+		if ( ! $this->is_null_allowed( $parameter_args ) ) {
+			return false === $this->phpcsFile->findNext(
+				\T_NULL,
+				$parameters[ $position ]['start'],
+				( $parameters[ $position ]['end'] + 1 )
+			);
+		}
+
+		return true;
+	}
+
+	/**
+	 * Determines if null is allowed for the specific function argument.
+	 *
+	 * @param array $parameter_args The arguments of the parameter supplied to the function.
+	 *
+	 * @return bool True if null is allowed, False otherwise.
+	 */
+	protected function is_null_allowed( $parameter_args ) {
+		if ( ! isset( $parameter_args['allow_null'] ) ) {
+			return false;
+		}
+
+		return $parameter_args['allow_null'];
+	}
+}
--- a/Yoast/Tests/WP/RequiredOptionalParametersUnitTest.php
+++ b/Yoast/Tests/WP/RequiredOptionalParametersUnitTest.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Unit test class for the Yoast Coding Standard.
+ *
+ * @package Yoast\YoastCS
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace YoastCS\Yoast\Tests\WP;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+/**
+ * Unit test class for the RequiredOptionalParameters sniff.
+ *
+ * @package Yoast\YoastCS
+ *
+ * @since   1.1.0
+ */
+class RequiredOptionalParametersUnitTest extends AbstractSniffUnitTest {
+
+	/**
+	 * Returns the lines where errors should occur.
+	 *
+	 * @return array <int line number> => <int number of errors>
+	 */
+	public function getErrorList() {
+		return array(
+			10 => 1,
+			13 => 1,
+			16 => 1,
+			22 => 1,
+			24 => 1,
+		);
+	} // end getErrorList()
+
+	/**
+	 * Returns the lines where warnings should occur.
+	 *
+	 * The key of the array should represent the line number and the value
+	 * should represent the number of warnings that should occur on that line.
+	 *
+	 * @return array<int, int>
+	 */
+	public function getWarningList() {
+		return array(
+			4  => 1,
+		);
+	}
+}
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,6 +15,7 @@
 /phpunit.xml export-ignore
 /phpunit.xml.dist export-ignore
 /Yoast/Tests export-ignore
+/Tests export-ignore
 
 #
 # Auto detect text files and perform LF normalization
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,17 +24,18 @@ php:
     - nightly
 
 env:
-  # Test against the highest supported PHPCS version.
-  - PHPCS_BRANCH="dev-master" PHPLINT=1
-  # Test against the lowest supported PHPCS version.
-  - PHPCS_BRANCH="3.3.1"
+  # Test against the highest and lowest supported PHPCS and WPCS versions.
+  - PHPCS_BRANCH="dev-master" WPCS_BRANCH="dev-develop" PHPLINT=1
+  - PHPCS_BRANCH="3.3.2" WPCS_BRANCH="dev-develop"
+  - PHPCS_BRANCH="dev-master" WPCS_BRANCH="1.2.0"
+  - PHPCS_BRANCH="3.3.2" WPCS_BRANCH="1.2.0"
 
 matrix:
   fast_finish: true
   include:
     # Extra build to check for XML codestyle.
     - php: 7.1
-      env: PHPCS_BRANCH="dev-master" SNIFF=1
+      env: PHPCS_BRANCH="dev-master" WPCS_BRANCH="dev-master" SNIFF=1
       addons:
           apt:
               packages:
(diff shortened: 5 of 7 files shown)


jrfnl:
> do I need to give props for the inspirational files and extended class and if so, how?

Just saying "Thank you" to me will do ;-)

Other than that, I'll have a look & review the sniff in a little. ~~Before I do - any particular reason not to sniff for `add_option` as I suggested in the ticket ?~~ Scratch that, just noticed you added a new commit adding it.

moorscode:
I was too wrapped up in getting the context that I didn't even read your comment on the issue until I created the PR.

Have read it now and added commits accordingly 😊

Tests/bootstrap.php
```diff
@@ -0,0 +1,64 @@
...
+ * Bootstrap file for running the tests on PHPCS 3.x.
+ *
+ * Load the PHPCS autoloader and the WPCS PHPCS cross-version helpers.
+ *
+ * {@internal We need to load the PHPCS autoloader first, so as to allow their
+ * auto-loader to find the classes we want to alias for PHPCS 3.x.
+ * This aliasing has to be done before any of the test classes are loaded by the
+ * PHPCS native unit test suite to prevent fatal errors.
+ *
+ * @package WPCS\WordPressCodingStandards
```
jrfnl: I think this docblock still needs adjusting, credit should be given for the file to WPCS, but the package, link and since tags should be for YoastCS.

Tests/bootstrap.php
```diff
@@ -0,0 +1,64 @@
...
+
+Please read the contributors guidelines for more information:
+https://is.gd/contributing2WPCS
+';
+
+	die( 1 );
+}
+
+// Load our class aliases.
+require_once dirname( __DIR__ ) . $ds . 'autoload-bootstrap.php';
```
jrfnl: The path to the WP alias file is different in a test situation, compared to a "run" situation, so loading the ruleset bootstrap here won't work as the paths won't match, so you need to load the alias file directly.

```php
require_once dirname( __DIR__ ) . '/vendor/wp-coding-standards/wpcs/WordPress/PHPCSAliases.php';
```

In a test situation, YoastCS is the root.
In a "run" situation, YoastCS is installed in the vendor directory of a project.

Yoast/Sniffs/Files/FileNameSniff.php
```diff
@@ -11,7 +11,6 @@
 
 use PHP_CodeSniffer\Sniffs\Sniff;
 use PHP_CodeSniffer\Files\File;
-use PHP_CodeSniffer\Util\Tokens;
```
jrfnl: This change does not belong in this PR.

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,93 @@
...
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+			// Check that number of parameters defined is not less than the position to check.
+			if ( isset( $parameters[ $position ] ) ) {
+				continue;
+			}
+
+			$message = '%s() is expected to be called with the "%s" argument at #%s.';
```
jrfnl: This sentence does not read right. Might help to change `at #%s` to `at position %d`.

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,93 @@
...
+			$code    = $this->string_to_errorcode( $matched_content . ucfirst( $parameter_args['name'] ) . 'ParamMissing' );
+
+			$data = array(
+				$matched_content,
+				$parameter_args['name'],
+				$position,
+			);
+
+			$this->addMessage( $message, $stackPtr, $parameter_args['enforce'], $code, $data, 0 );
```
jrfnl: `$severity` is optional and should normally not be set, so please remove it.

Yoast/Tests/WP/RequiredParametersUnitTest.inc
```diff
@@ -0,0 +1,16 @@
+<?php
+
+// Missing autoload option.
+update_option( 'a', 'b' );
+
```
jrfnl: Maybe also add a test case where `null` is passed as the third param ? That's the WP default value after all (for `update_option()`).

And what about when the param has been set to `false`/ `no` ?

Yoast/Tests/WP/RequiredParametersUnitTest.php
```diff
@@ -0,0 +1,47 @@
...
+ * @package Yoast\YoastCS
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace YoastCS\Yoast\Tests\WP;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+/**
+ * Unit test class for the FileName sniff.
```
jrfnl: Needs adjusting.

Tests/bootstrap.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Bootstrap file for running the tests on PHPCS 3.x.
```
jrfnl: `on PHPCS 3.x` is superfluous as YoastCS only supports PHPCS 3.x

Tests/bootstrap.php
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Load the PHPCS autoloader and the WPCS PHPCS cross-version helpers.
+ *
+ * {@internal We need to load the PHPCS autoloader first, so as to allow their
```
jrfnl: This comment is not needed in the context of the YoastCS standard as YoastCS does not need the aliasing as such.
It could be changed to refer to the fact that WPCS at this point in time still needs the aliases and that as you extend a WPCS class, it is needed here too.

As a side-note: WPCS will be dropping PHPCS 3.x support by the end of the summer, so at that time, the aliases will be removed from WPCS completely.

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,93 @@
...
+ * @author  Jip Moors
+ *
+ * @since   0.6
+ */
+class RequiredParametersSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * Intended to be overruled in the child class.
```
jrfnl: Comment line is superfluous as this *is* a child class.

Yoast/ruleset.xml
```diff
@@ -2,6 +2,9 @@
 <ruleset name="Yoast" namespace="YoastCS\Yoast">
 	<description>Yoast Coding Standards</description>
 
+	<!-- Autoload the autoloaders of the dependencies. -->
```
jrfnl: I've ran some tests, but I don't think this is actually needed as the ruleset includes the WPCS `WordPress` ruleset and that ruleset already loads the WPCS autoload. This can be (and should be) removed.

autoload-bootstrap.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
jrfnl: As per the comment about autoload in the ruleset + the comment about the paths being different in a test situation and needing to include the WPCS alias file directly for a test situation, this whole file isn't necessary.

autoload-bootstrap.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Autoload the autoload files of the individual standards.
+ *
+ * @category PHP
+ * @package  YoastCS
+ * @author   Jip Moors
+ */
+
+if ( file_exists( __DIR__ . '/vendor/wimg/php-compatibility/PHPCSAliases.php' ) ) {
```
jrfnl: Side-note/Teaching moment: the PHPCompatibility alias file would not be needed anyway as - again - their own ruleset loads it.

Basically, the only time when you need to load the alias files of another standard is when you just load sniffs from another standard, not the standard itself.

So, for the QA-WP project, I load them as those rulesets don't include the whole WP standard, but only individual sniffs.
For the unit tests you need to load them as the sniffs are tested individually, independent of the ruleset.

jrfnl (changes requested):
Reviewed. Besides the remarks I've left inline, I'm wondering what the goal is of the checks currently in the sniff.

Is your intention that the param should always be set ?
Or that the param is always set to `true`/`yes` ?

The default is `true`/`yes` for `add_option()` and whatever was set when the option was originally added for `update_option()`, defaulting to `true`/`yes` when the option doesn't exist yet.

So what are you really trying to archive with this sniff ?

jrfnl:
Oh.. and two more things:
* The `Tests` directory also needs to be added to the `gitattributes` `export-ignore` section.
* The Travis script needs adjusting as the `phpunit` command now adds a `bootstrap` on the command-line which overrules the one you've added.

moorscode:
The main goal of the Sniff is to make sure the `autoload` is set according to intend. We want to avoid having options autoloaded if they are not needed outside of the admin.

Having the default `yes` is not wanted for most of the situations where options are used.

@jdevalk has a ticket in progress/in mind where the level of options will be extended with `autoload for admin`, to make sure we can still autoload everything that is needed efficiently in the backend, but also loading only what is actually needed on the frontend.

Yoast/Sniffs/Files/FileNameSniff.php
moorscode: Technically correct. Will it stop you from merging the PR?

moorscode:
I've processed all the feedback.

Tests/bootstrap.php
```diff
@@ -0,0 +1,82 @@
...
+
+if ( ! defined( 'PHP_CODESNIFFER_IN_TESTS' ) ) {
+	define( 'PHP_CODESNIFFER_IN_TESTS', true );
+}
+
+$ds = DIRECTORY_SEPARATOR;
+
+// Get the PHPCS dir from an environment variable.
+$phpcsDir = getenv( 'PHPCS_DIR' );
+$phpcsComposerDir = implode( $ds, array( dirname( __DIR__ ), 'vendor', 'squizlabs', 'php_codesniffer' ) );
```
jrfnl: The implode logic feels like something which should be abstracted to a helper function as it's used more than once.

Aside from that, it may be useful to set a `$phpcsComposerBaseDir = dirname( __DIR__ );` above this line.

Tests/bootstrap.php
```diff
@@ -0,0 +1,82 @@
...
+	/*
+	 * As of PHPCS 3.1, PHPCS support PHPUnit 6.x, but needs a bootstrap, so
+	 * load it if it's available.
+	 */
+	if ( file_exists( $phpcsDir . $ds . 'tests' . $ds . 'bootstrap.php' ) ) {
+		require_once $phpcsDir . $ds . 'tests' . $ds . 'bootstrap.php';
+	}
+}
+else {
+	echo 'Uh oh... can\'t find PHPCS. Are you sure you are using PHPCS 3.x ?
```
jrfnl: This second part of this line is in WPCS as it still supports both PHPCS 2.x as well as 3.x. It is not needed in YoastCS.

Tests/bootstrap.php
```diff
@@ -0,0 +1,82 @@
...
+If you use Composer, please run `composer install`.
+Otherwise, make sure you set a `PHPCS_DIR` environment variable in your phpunit.xml file
+pointing to the PHPCS directory.
+';
+
+	die( 1 );
+}
+
+// Get the WPCS dir from an environment variable.
+$wpcsDir = getenv( 'WPCS_DIR' );
```
jrfnl: :+1:

Tests/bootstrap.php
```diff
@@ -0,0 +1,82 @@
...
+
+// This may be a Composer install.
+if ( false === $wpcsDir && is_dir( $wpcsComposerDir ) ) {
+	$wpcsDir = $wpcsComposerDir;
+}
+elseif ( false !== $wpcsDir ) {
+	$wpcsDir = realpath( $wpcsDir );
+}
+
+// Try and load the PHPCS autoloader.
```
jrfnl: `PHPCS` -> `WPCS`

Yoast/Sniffs/Files/FileNameSniff.php
jrfnl: It will not ;-)

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,91 @@
+<?php
+/**
+ * YoastCS\Yoast\Sniffs\Files\FileNameSniff.
```
jrfnl: Sorry, I must have missed this the first time. Wrong class name mentioned here.

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,91 @@
...
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+			// Check that number of parameters defined is not less than the position to check.
+			if ( isset( $parameters[ $position ] ) ) {
+				continue;
+			}
+
+			$message = '%s() is expected to be called with the "%s" argument at the %d position.';
```
jrfnl: `at the 3 position` is still grammatically incorrect. `at position %d` is grammatically correct.

jrfnl: And just wondering - `is expected to` sounds like a `warning`, not an `error`, while depending on context you throw either an `error` or a `warning`.

Should this phrase be changed depending on the `error`/`warning` context to sound stricter ? `%s() must be called with the ...`

Yoast/Tests/WP/RequiredParametersUnitTest.inc
```diff
@@ -0,0 +1,40 @@
...
+add_option( 'a', 'b', '' );
+
+// Ok.
+add_option( 'a', 'b', '', 'yes' );
+
+// Ok.
+add_option( 'a', 'b', '', 'no' );
+
+// Ok.
+add_option( 'a', 'b', '', null );
```
jrfnl: `null` is not a valid value to pass to `add_option()`. See my remark in my overall review comment.

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,91 @@
...
+ * @author  Jip Moors
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace YoastCS\Yoast\Sniffs\WP;
+
+use WordPress\AbstractFunctionParameterSniff;
+
+/**
+ * Verifies that optional parameters are set, when they should be set.
```
jrfnl: `, when they should be set` is redundant. These parameters are _optional_ as stated in the first part of the sentence, so there is no _should be set_, other than that the YoastCS standard wants them to be, which is what the sniffs is all about in the first place.

Yoast/Sniffs/WP/RequiredParametersSniff.php
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @package Yoast\YoastCS
+ * @author  Jip Moors
+ *
+ * @since   0.6
+ */
+class RequiredParametersSniff extends AbstractFunctionParameterSniff {
```
jrfnl: Just wondering - would `RequiredOptionalParameters` better describe what the sniff covers ?

Tests/bootstrap.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * Bootstrap file for running the tests.
+ *
+ * Load the PHPCS autoloader and the WPCS PHPCS cross-version helpers.
```
jrfnl, line 5: You asked about giving credit, but don't mention that the bulk of the code in this file comes from/is inspired by WPCS at all now....

jrfnl (changes requested):
@moorscode Looking good. Nearly there, though I still have a couple of remarks:

### 1. Travis script

The Travis script was set up to:
- pull in PHPCS at a `PHPCS_BRANCH` version, set in the matrix, so the unit tests would be run against the minimum and maximum supported PHPCS version.
- pull in WPCS and PHPCompatibility for the CS check of the YoastCS code only, so using the `master` branch was fine.

However, now you need WPCS (not PHPCompatibility) for unit testing the sniffs too.
But the sniffs will normally be installed using Composer, which fixes WPCS at a certain version range, so using the WPCS `master` branch does not guarantee that the sniffs are unit tested based on the code which the "end-users" will use.

In other words: as the version of WPCS has now become important, the Travis script needs to be changed. It has to pass the exact branch(es) needed for WPCS to the `git clone` command.

In the future, the matrix might also need to be extended to test against the minimum and maximum supported version of WPCS if more than one version is supported.
At this time, WPCS `0.14.0` and `0.14.1` are supported, but as `0.14.1` contains no significant changes compared to `0.14.0` for the abstract sniff you are using, just testing against `0.14.1` will be fine.

You could also consider switching over to using Composer to install the dependencies, but there are some snakes under the grass with that so if you want, I could pull that change separately later.

### 2. The parameter should be set

> The main goal of the Sniff is to make sure the autoload is set according to intend.

If that's the case, an explicit value of `null` for the parameter should not be accepted, as the parameter in that case is still not set.

To fix this, you can test the `raw` value of the parameter, so I would suggest changing line 75 of the sniff:
```php
if ( isset( $parameters[ $position ] ) ) {
```

to:
```php
if ( isset( $parameters[ $position ] ) && $parameters[ $position ]['raw'] !== 'null' ) {
```


### 3. Unit tests

My previous remark `And what about when the param has been set to false/ no ?` was not intended to say that every single possible value needs to be tested.
I was just looking to see some more variety in the unit tests.

Testing every possible value is just testing the same thing several times.

For this particular group of functions, I think testing each function with each type of value, i.e. no param, `null`, a boolean and a string, should be sufficient.
Between those test cases, you can then have the suggested variety by having one of the functions use `true`, the other `false` etc.


### Other

Will you clean up the branch to topical commits ? Or shall I squash on merge ?

moorscode:
> Will you clean up the branch to topical commits ? Or shall I squash on merge ?
I would love to clean up the branch, but I'm missing the context on the specific tools and commands to do so at this point.

I'd say a squash would be fine for this one.
I think having a dev-presentation about achieving this would be very nice (perhaps even have it recorded for easy playback).

Tests/bootstrap.php
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Code has been inspired on the phpcs3-bootstrap.php used in the WordPress Coding Standards project
```
jrfnl: Nitpick: `inspired by` or `based on`. _Inspired on_ is grammatically incorrect.

Nitpick 2: The `see` keyword/tag in docblocks is intended to be used for FQCN, the `@link` tag for URLs. And for documentation purposes, linking to the actual file is better as now another dev would still need to dig through the WPCS repo to find the file.

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,120 @@
...
+
+	/**
+	 * Functions this sniff is looking for.
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *    $target_functions = array(
+	 *        (string) Function name. => array(
+	 *            (int) Target parameter position, 1-based. => array(
+	 *                'name'    => (string)  The name of the argument that should be set.
+	 *                'enforce' => (boolean) If this should trigger an error or warning.
```
jrfnl: Documentation for `allow_null` missing.

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,120 @@
...
+			),
+		),
+		'add_option'    => array(
+			4 => array(
+				'name'       => 'autoload',
+				'enforce'    => true,
+				'allow_null' => false,
+			),
+		),
+	); // End $target_functions.
```
jrfnl: These type of `// End` comments are not needed/required. Is there a particular reason why you are adding it ?

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,120 @@
...
+
+	/**
+	 * Process the parameters of a matched function.
+	 *
+	 * This method has to be made concrete in child classes.
```
jrfnl: Superfluous comment line: You are in the child class.

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,120 @@
...
+			if ( ! $is_error && ! $parameter_args['allow_null'] ) {
+				$is_error = ( isset( $parameters[ $position ] ) && $parameters[ $position ]['raw'] === 'null' );
+			}
+
+			$this->addMessage( $message, $stackPtr, $is_error, $code, $data );
+		}
+	}
+
+	/**
+	 * Determines if the parameter provided is valid, if any.
```
jrfnl: Grammatically better: "_Determines if a valid parameter has been provided._"

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,120 @@
...
+	 *
+	 * @param array $parameters     The parameters used to call the function.
+	 * @param int   $position       The position of the parameter to check.
+	 * @param array $parameter_args The configuration of the function check.
+	 *
+	 * @return bool True if a valid parameter has been given.
```
jrfnl: :heavy_plus_sign:  `False otherwise.`

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,120 @@
...
+	 */
+	protected function has_valid_parameter( $parameters, $position, $parameter_args ) {
+		if ( ! isset( $parameters[ $position ] ) ) {
+			return false;
+		}
+
+		if ( ! $parameter_args['allow_null'] && $parameters[ $position ]['raw'] === 'null' ) {
```
jrfnl: Will this check always need to be executed ? In other words: will `allow_null` always be set ?

Yoast/Tests/WP/RequiredOptionalParametersUnitTest.php
```diff
@@ -0,0 +1,49 @@
...
+ * @package Yoast\YoastCS
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace YoastCS\Yoast\Tests\WP;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+/**
+ * Unit test class for the RequiredParameters sniff.
```
jrfnl: `RequiredParameters` =>`RequiredOptionalParameters`

jrfnl (changes requested):
@moorscode We're very nearly there. Looking good indeed :+1:, just some dotting of the i's and crossing of the t's left.

There is just one last issue I'd like to discuss based on the last round of changes:

You currently have one error message with one error code being thrown from this sniff.
This message can be either a warning or an error and the content of the message changes based on two, three different things.

The idea behind the error codes is that each code identifies a unique message. This enables modular disabling/enabling of sniffs and codes, both from the ruleset as well as inline.

The current message does not comply with that as it basically has four different messages folded into one, all using the same error code.

So, what about actually having two different errors and each having two different codes based on their error/warning status ?

I'm thinking along the lines of:
1. `%s() is %s to be called with the "%s" argument at position %d%s.` with depending on error/warning status a different error code, something along the lines of `ParamRequired`/`ParamExpected`/`ParamMissing`.
    Take your pick or come up with your own better alternative.
2. `Passing null to the "%s" argument of %s() is %s` where the last `%s` would expand to `forbidden` or `discouraged`.
    For the error code you could then, for instance, use `NullForbidden` and `NullDiscouraged`.


Other than that, some practical points:
* :+1: I will squash the commits. Let's plan that dev session about having a clean & useful git history via Slack.
* Would you like me to create a PR to change the Travis file to use Composer ? As per https://github.com/Yoast/yoastcs/pull/76#pullrequestreview-108497845
* You are rightfully using version `0.6.0` for this sniff, as the introduction of a new sniff should up the version minor.
   There have been some PRs merged between the `0.5.0` tag and now. Should `0.5.1` be tagged before this PR is merged ? Or should those other merged PRs be retagged as `0.6.x` ?

Tests/bootstrap.php
moorscode: Can you have multiple `@link` tags in a documentation block? As there already is one referencing to the project itself a couple of lines down.

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
moorscode: It was present in the file I copied it from (the abstract class this is extending).

moorscode:
Push failed because I intended to create a new branch to rewrite the history of the commits and pushed the changes on that branch. I have deleted the branch in the meanwhile, causing the Travis builds to fail because they can no longer check out that branch.

Tests/bootstrap.php
jrfnl: You can, but more than that, you can use inline `@link` tags, which in this case would probably be more appropriate. `{@link URL link text}` and PHPDocumentor will handle it perfectly ;-)

See: https://docs.phpdoc.org/references/phpdoc/tags/link.html

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
jrfnl: The abstract class doesn't actually have the end tag. I think you may have copied it from the `WP.DeprecatedParameters` sniff, but that's a different situation as the array there is nearly 200 lines long.

moorscode:
I have split up the error and warning messages to have their own identifiers.
Rephrased the error codes to be more readable and include `Null` if disallowed.

Examples:
- "update_optionOptionalParamAutoloadMissingNullDisallowed"
- "add_optionRequiredParamAutoloadMissingNullDisallowed"

jrfnl:
FYI - the Travis script change over to use Composer as discussed in https://github.com/Yoast/yoastcs/pull/76#pullrequestreview-108497845, is included in PR #81.

Yoast/Tests/WP/RequiredOptionalParametersUnitTest.php
```diff
@@ -0,0 +1,50 @@
...
+	 */
+	public function getErrorList() {
+		return array(
+			10 => 1,
+			13 => 1,
+			16 => 1,
+			22 => 1,
+			24 => 1,
+		);
+	} // end getErrorList()
```
jrfnl, line 35: End comments are not needed.

Yoast/Tests/WP/RequiredOptionalParametersUnitTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+	/**
+	 * Returns the lines where warnings should occur.
+	 *
+	 * The key of the array should represent the line number and the value
```
jrfnl, line 40: This "long description" can be removed (as it has been in the method above and in all the other test files).

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,146 @@
...
+ * @author  Jip Moors
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace YoastCS\Yoast\Sniffs\WP;
+
+use WordPress\AbstractFunctionParameterSniff;
+
+/**
+ * Sniff that ensures the optional parameter for update_option and set_option
```
jrfnl:
```suggestion
 * Sniff that ensures the optional parameter for update_option() and add_option()
```

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,146 @@
...
+ * are defined because we want developers to be conscious about what is needed.
```
jrfnl:
```suggestion
 * are passed because we want developers to be conscious about what is needed.
```

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,146 @@
...
+	/**
+	 * Functions this sniff is looking for.
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *    $target_functions = array(
+	 *        (string) Function name. => array(
+	 *            (int) Target parameter position, 1-based. => array(
+	 *                'name'       => (string)  The name of the argument that should be set.
+	 *                'enforce'    => (boolean) If this should trigger an error or warning.
+	 *                'allow_null' => (boolean) If `null` is seen as setting the optional value, defaults to False.
```
jrfnl:
```suggestion
	 *                'allow_null' => (boolean) Whether `null` is seen as setting the optional value. Defaults to false.
```

Yoast/Sniffs/WP/RequiredOptionalParametersSniff.php
```diff
@@ -0,0 +1,146 @@
...
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+			// Check that number of parameters defined is not less than the position to check.
```
jrfnl:
```suggestion
			// Check that a valid parameter value is being passed.
```

jrfnl (commented):
@moorscode I've reviewed the sniff again and left some (minor) inline remarks.

Aside from that, a couple of generic points I'd like to discuss with you when you have the time:
* The way the logic is written now, the mental overhead to read the code is quite high.
* Even though, the errorcode is now modular, IMO you are trying to do to much with one error.
    Why not have two separate checks ? One that the parameter is set and a second that it's not `null` ? and have separate errors for each.
* Both errors and warnings should be solvable by other means that adding an `phpcs:ignore` comment.
    If an option is being added with `add_option()` with the `$autoload` parameter set and in another part of the code that same option is updated using `update_option()` without setting the `$autoload` parameter - to prevent overruling whatever was set with `add_option()` -, there is currently no easy way to "fix" this in the code, other than duplicating the `$autoload` parameter used in the original `add_option()` call and creating a potential maintenance/debug nightmare where every single `add_option()` and `update_option()` call for each option has to be kept in line with each other and if the `$autoload` is changed for one, it has to be changed in every single `update_option()` call for that same option.
    I think this needs to be thought through a little more before this sniff can be merged.

jrfnl:
Another point which will need looking into - now or later (non-urgent): the unit tests against PHP `nightly` (7.4) are failing specifically on this sniff.
See: https://travis-ci.org/Yoast/yoastcs/builds/454169892

I'm not sure if this is PHPCS. WPCS or YoastCS related as I haven't looked into it.
Note: both the PHPCS native tests as well as the WPCS native tests currently pass against `nightly`.

moorscode:
> If an option is being added with add_option() with the $autoload parameter set and in another part of the code that same option is updated using update_option() without setting the $autoload parameter - to prevent overruling whatever was set with add_option() -, there is currently no easy way to "fix" this in the code, other than duplicating the $autoload parameter used in the original add_option() call and creating a potential maintenance/debug nightmare where every single add_option() and update_option() call for each option has to be kept in line with each other and if the $autoload is changed for one, it has to be changed in every single update_option() call for that same option.
I think this needs to be thought through a little more before this sniff can be merged.

These `update_option` and `add_option` functions are a bit silly looking at the code.

As the `update_option` calls `add_option` when it does not exist, and `add_option` tries to INSERT the option into the table but does an UPDATE when the key already exists..

We have only one usage of `add_option` in our code and thinking about it I would rather make sure all calls are just `update_option` requests with the enforced `autoload` argument.

If the same option is updated on multiple locations the autoload context must be thought about anyway. If we change if something needs to be autoloaded doing a search would be logical to determine where it is actually set/used.

/cc @atimmer

jrfnl:
@moorscode So.. how about we try and prevent this PR having a one-year anniversary ? Did you have a chance to think over the things we discussed in December ?

PR #79, opened 2018-06-08
Update for phpcs 3.3.0

jrfnl:
### Use PHP_CodeSniffer v 3.3.0

This updates YoastCS to use PHPCS 3.3.0 which was released this week.
See: https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.3.0

The most significant change in the context of YoastCS is that `array` type properties set in the ruleset now use a new format. The old format is still supported for now, but will be removed in PHPCS 4.0.

Other than that, this version contains lots of bug fixes so updating is a good idea.

#### Revert "Travis: fix build failure"

As PHPCS 3.3.0 supports PHPUnit 7.x, there is no need for this fix anymore. This reverts commit 9be33ae.

### Ruleset: add two new sniffs

The `LowerCaseType` sniff verifies that all type casts, parameter and return type declarations are in lowercase (with the exception of classname based type declarations).
The `ShortFormTypeKeywords` sniff ensures that the same type indicators are used in short form, i.e. `int`, not `integer`.

I don't expect any new errors to be thrown up because of these sniffs as what they check is common practice, but, just in case, it will now be enforced.

These sniffs are new to PHPCS 3.3.0 and will be included in WPCS once the WPCS minimum PHPCS version will be upped to PHPCS 3.3.0. This is expected to happen in WPCS 2.0. So, once YoastCS moves onto WPCS 2.0, these additions can be removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ dist: trusty
 
 cache:
     apt: true
-    directories:
-      - tmp/phpunit
-
 
 php:
     - 5.4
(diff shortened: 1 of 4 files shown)


Dieterrr:
CR done 👍

PR #88, opened 2018-08-06
New sniff to check that test double/mock classes are in the correct directory

### LINKED ISSUE #75: Create a sniff to make sure test doubles/mocks are in a `doubles` directory
To determine whether a sniff is a double/mock, there are a couple of things which can be checked:
1. Does the class (interface/trait) name have `double`/`mock` in it ?
2. Does the class extend one of the known unit test classes ? If not and it *is* a file in the `test(s)` directory, this may be a helper class or an incorrectly named mock/double class.
    If so, this should probably trigger an error/warning too.

Next would be a check that those class files reside in a `tests/doubles/` directory. However, this path should be one which can be overruled by a property in the ruleset to allow - for instance - the premium plugin, to have their mock/doubles in a `tests/premium/doubles/` directory.

I have the bare bones of this sniff ready already, but there are some open questions (most notably see point 2 above) and it will need clean up & unit tests.

jrfnl:
This sniff looks for classes containing `Mock` or `Double` (case-insensitive) in the class name and will:
1. Check that a `tests/doubles` directory exist and throw an `error` if it doesn't.
2. Check that the file containing the double/mock class is within the `tests/doubles` directory and throw an `error` if it isn't.
3. Check that the double/mock class is in a file of its own and throw an `error` if it isn't.

The `doubles_path` by default is `/tests/doubles` relative to the `basepath` for the project.
This can be customized from within a custom ruleset by passing a different value for `doubles_path`.
```xml
<rule ref="Yoast.Files.TestDoubles">
	<properties>
		<property name="doubles_path" value="/tests/assets"/>
	</properties>
</rule>
```

For this sniff to function, the `basepath` for the project has to be set. If it's not a `warning` will be thrown.
The basepath will normally be set from within the custom ruleset like so:
```xml
<arg name="basepath" value="./"/>
```

Includes unit tests covering all errors and warnings.

As YoastCS itself does a) not contain mocks/doubles and b) has to follow the predetermined directory structure as set by PHPCS, for the YoastCS native CS check, this sniff is disregarded (excluded).

Fixes #75

----

Potential changes which could be made to tighten the sniff up against false positives:
* Limit the matching of the class name to `Double` / `Mock` at the _end_ of the class name.
* Limit the checking for this issue to classes found in a `tests` directory.

AFAICS, at this moment, no false positives are thrown without these changes and issues which previously existed and which this sniff warns about, have already been solved for the plugins (based on scans using an earlier version of this sniff).
If at some point in the future false positives would be thrown because of the above, the additional checks could be build in.


moorscode:
CR Done 👍

PR #90, opened 2018-08-23
Add a changelog for the YoastCS package

### LINKED ISSUE #77: Add a changelog
As there is more active development going on and expected in the near future for this repo, it may be advisable to add a `changelog.md` file (or add it to the readme).

Once the initial changelog is made, tagged releases should be updated with the relevant change info.

jrfnl:
A well-maintained changelog makes it easier for the "product-owners" of each Yoast plugin to upgrade to a higher version of YoastCS, as it is documented what has changed in each version.

As no changelog existed until now, this PR adds a changelog for all YoastCS releases since the very first one.

Fixes #77

N.B.: This PR includes the changelog for the upcoming YoastCS 1.0.0 release.

To do:

- [x] Add the changelogs for each release to the GH release tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,199 @@
+# Change Log for YoastCS
+
+All notable changes to this project will be documented in this file.
+
+This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a CHANGELOG](https://keepachangelog.com/).
+
+
+### [1.0.0] - 2018-08-24
+
+#### Added
+* PHPCS: New `Yoast.Files.TestDoubles` sniff.
+    This sniff verifies that test `double`/`mock` classes are in their own file in a `doubles` test sub-directory.
+* PHPCS: New `Yoast.WhiteSpace.FunctionSpacing` sniff.
+    This sniff is based on the PHPCS native `Squiz.WhiteSpace.FunctionSpacing` sniff and verifies and auto-fixes the amount of blank lines between methods within OO-structures.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,199 @@
...
+* Composer: Supported version of [PHP_CodeSniffer] has been changed from `^2.2.0` to `2.5.1`.
+* Composer: Supported version of [WordPressCS] has been changed from `0.6` to `0.9`.
+* Composer: Included version of [PHP Mess Detector] has been updated from `2.2.3` to `2.4.2`.
+    The minimum supported version remains at `2.2.3`.
+
+#### Removed
+* PHPCS: The exclusion of the whole `WordPress.NamingConventions.ValidFunctionName` sniff from the YoastCS ruleset.
+
+
+### 0.1 - 2016-01-08
```
moorscode, line 180: Missing square brackets around version number.

moorscode:
Do we need to create Github releases, or should we just keep the release tags? @atimmer

CHANGELOG.md
jrfnl: Actually, they are not missing :smiley_cat: 

The square brackets indicate a link and the release numbers all link to the commit summary, i.e. a compare between the previous tag and the release tag.

As version `0.1` has no previous tag, there is nothing which can be linked to.

jrfnl:
> Do we need to create Github releases, or should we just keep the release tags?

What with the changelog already being available, this would now just be a case of copy/pasting, which is why I mentioned it in the original ticket above as an action.

CHANGELOG.md
moorscode: Do'h! Now it all makes sense /facepalm

PR #93, opened 2018-09-11
Add more files to `.gitignore`

jrfnl:
These files are used locally by developers to override the `.dist` version of the same files as contained in this repo and should not be committed to the repo.


IreneStr:
CR 👍

PR #96, opened 2018-09-20
New `CodeCoverageIgnoreDeprecated` sniff

### LINKED ISSUE #95: Suggestion: sniff to check that deprecated methods are ignored for code coverage
Related to https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/993 which proposes a sniff to check various other aspects of deprecated functions/methods.

In addition to the WPCS sniff, I would like to suggest a sniff to check that any function/method which has a `@deprecated` tag in the Yoast plugins, also has a `@codeCoverageIgnore` tag.

Files which are in a `deprecated` folder could be ignored for this sniff (presuming that the PHPUnit code coverage configuration ignores those anyway).

Opinions ?

jrfnl:
This sniff will check that functions which have a `@deprecated` tag in the function docblock, also have a  `@codeCoverageIgnore` tag in the same docblock.

If a `codeCoverageIgnore` text is found without the `@` sign, it will be recognized by the sniff and can be auto-fixed.

Individual plugins may have certain files `exclude`d for codeCoverage in the PHPUnit configuration file.
For those plugins corresponding `exclude`s need to be added to the PHPCS configuration file to ignore those files completely for the purposes of this sniff.

Example:
```xml
<rule ref="Yoast.Commenting.CodeCoverageIgnoreDeprecated">
    <exclude-pattern>*/deprecated/*\.php$</exclude-pattern>
</rule>
```

Fixes #95

### DIFF
--- a/Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc
+++ b/Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc
@@ -1,0 +1,56 @@
+<?php
+
+class SomeThing {
+
+	/**
+	 * Active (non-deprecated) function.
+	 */
+	function activeFunction() {}
+
+	/**
+	 * @deprecated x.x
+	 * @codeCoverageIgnore
+	 */
+	public /* don't get confused by this */ function correctDocblock() {
+		// Code
+    } // end function
+
+    public function functionWithoutDocblock {}
+
+	/**
+	 * @codeCoverageIgnore
+	 * @deprecated x.x
+	 */
+	private static function correctDocblockReverseOrder() {}
+
+	/**
+	 * @deprecated x.x
+	 *
+	 * @param string $param1 Short desc.
+	 * @param string $param2 Short desc.
+	 *
+	 * @return bool
+	 *
+	 * @codeCoverageIgnore
+	 */
+	protected function correctDocblockOtherTagsIrrelevant() {}
+
+	/**
+	 * @deprecated x.x
+	 *
+	 * codeCoverageIgnore
+	 */
+	private static function ignoreNotAsTag() {}
+
+	/**
+	 * @deprecated x.x
+	 *
+	 * This is a sentence containing the phrase codeCoverageIgnore, but should not be regarded as the tag.
+	 */
+	static protected function missingCodeCoverageIgnore() {}
+
+	/**
+	 * @deprecated x.x
+	 */
+	public function missingCodeCoverageIgnore() {}
+}
--- a/Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc.fixed
+++ b/Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc.fixed
@@ -1,0 +1,56 @@
+<?php
+
+class SomeThing {
+
+	/**
+	 * Active (non-deprecated) function.
+	 */
+	function activeFunction() {}
+
+	/**
+	 * @deprecated x.x
+	 * @codeCoverageIgnore
+	 */
+	public /* don't get confused by this */ function correctDocblock() {
+		// Code
+    } // end function
+
+    public function functionWithoutDocblock {}
+
+	/**
+	 * @codeCoverageIgnore
+	 * @deprecated x.x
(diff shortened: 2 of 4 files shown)


moorscode:
Looks good to me 👍

Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	function activeFunction() {}
+
+	/**
+	 * @deprecated x.x
+	 * @codeCoverageIgnore
+	 */
+	public /* don't get confused by this */ function correctDocblock() {
+		// Code
+    } // end function
```
afercia, line 16: please notice this line uses spaces for indentation, it should use a tab character

Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc
```diff
@@ -0,0 +1,56 @@
...
+
+    public function functionWithoutDocblock {}
```
afercia, line 18: please notice this line uses spaces for indentation, it should use a tab character

Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc.fixed
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	function activeFunction() {}
+
+	/**
+	 * @deprecated x.x
+	 * @codeCoverageIgnore
+	 */
+	public /* don't get confused by this */ function correctDocblock() {
+		// Code
+    } // end function
```
afercia, line 16: please notice this line uses spaces for indentation, it should use a tab character

Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc.fixed
```diff
@@ -0,0 +1,56 @@
...
+
+    public function functionWithoutDocblock {}
```
afercia, line 18: please notice this line uses spaces for indentation, it should use a tab character

Yoast/Tests/Commenting/CodeCoverageIgnoreDeprecatedUnitTest.inc
jrfnl: @afercia Hi Andrea, FYI: that is the unit test case file for the sniff and the sniff should work style-independently, i.e. detect the issue it is targeting no matter what other CS issues are present.

So, while you are correct, in this case, it really doesn't matter and could be considered a feature.

afercia: @jrfnl hi! Thanks for the explanation. Hm then maybe consider an inline comment to explain it? Otherwise I'm afraid some pedant guy like me might inadvertently change them to tab, thinking to do the right thing 😄

jrfnl: That would leave the unit test case files full of inline comments. It might be a better idea to add a `CONTRIBUTING.md` file to this repos which includes a remark to that effect.
The unit test case files are easy to distinguish from real files as they have a `.inc(.fixed)` file extension and contain nonsense code.

@afercia Want to have a go at creating one ?

> Otherwise I'm afraid some pedant guy like me might inadvertently change them to tab, thinking to do the right thing

I look at all PRs here anyway (for now) and would reject/revert it if I came across it.

afercia: Sorry I'm afraid I don't have much time, pretty busy on a new ESLint config for _all_ the projects 😬

jrfnl: Very valuable as well. I'll make an issue for the Contributing file.

jrfnl: See #97

PR #108, opened 2018-11-12
TestDoubles: various improvements, including allowing multiple dirs

jrfnl:
As discussed with @moorscode, in rare circumstances, a plugin should be able to allow more than one directory to be used to place test doubles/mocks in.

This PR contains the sniff refactor to allow for that.

While looking at the sniff, I saw some more room for improvement, so I've fixed a few other things as well (in separate commits).

----

### TestDoubles: improve "OneObjectPerFile" check

While looking at this sniff, I realized that the "OneObjectPerFile" check only looked down, not up.

This commit fixes that.

### TestDoubles: stabilize slashes in base_path

Prevent the `$base_path` being double slashed if it would have a slash at the end of the path already.

### TestDoubles: improve path comparison

Make sure target_path always has an end slash to prevent a path like `/tests/doublesniff` to be recognized as correct when `/test/doubles` is in the allowed list.

This also changes the comparison from just using the `dirname` of the file being examined to using to complete file name, to prevent those kind of mismatches.

Unlikely to be a problem in real life, but the code should be correct.

### TestDoubles: allow for multiple paths

As discussed with @moorscode, in rare circumstances, a plugin should be able to allow more than one directory to be used to place test doubles/mocks in.

In this commit, the `doubles_path` sniff property is changed from a string to an array property.

In the previous version of this sniff, when the `doubles_path` would not be valid, both a `NoDoublesDirectory` as well as a `WrongDirectory` error would be thrown.

In this version, the `NoDoublesDirectory` error will only be thrown if none of the relative paths supplied exist.
If at least one path is valid and that path is not used, the sniff will only throw a `WrongDirectory` error.

Notes:
* Includes BC-break protection for if the value would still be supplied as a string to prevent BC-breaks.
    This snippet can be removed in YoastCS 2.0.0.
* Now uses PHPCS 3.2+ style annotations for the property changing within the test case files.
    This is preferred since setting array properties using `//phpcs:set` was greatly improved in PHPCS 3.3.0.
    See: squizlabs/PHP_CodeSniffer#1999

Includes unit tests.

### DIFF
--- a/Yoast/Sniffs/Files/TestDoublesSniff.php
+++ b/Yoast/Sniffs/Files/TestDoublesSniff.php
@@ -93,7 +93,7 @@ class TestDoublesSniff implements Sniff {
 
 		if ( ! isset( $this->target_path ) || defined( 'PHP_CODESNIFFER_IN_TESTS' ) ) {
 			$target_path  = $base_path;
-			$target_path .= ltrim( $this->normalize_directory_separators( $this->doubles_path ), '/' );
+			$target_path .= trim( $this->normalize_directory_separators( $this->doubles_path ), '/' ) . '/';
 
 			$this->target_path = false;
 			if ( file_exists( $target_path ) && is_dir( $target_path ) ) {
@@ -114,14 +114,9 @@ class TestDoublesSniff implements Sniff {
 			);
 		}
 
-		$path_info = pathinfo( $file );
-		if ( empty( $path_info['dirname'] ) ) {
-			return;
-		}
-
-		$tokens  = $phpcsFile->getTokens();
-		$dirname = $this->normalize_directory_separators( $path_info['dirname'] );
-		if ( false === $this->target_path || stripos( $dirname, $this->target_path ) === false ) {
+		$tokens       = $phpcsFile->getTokens();
+		$path_to_file = $this->normalize_directory_separators( $file );
+		if ( false === $this->target_path || stripos( $path_to_file, $this->target_path ) === false ) {
 			$phpcsFile->addError(
 				'Double/Mock test helper classes should be placed in the "%s" sub-directory. Found %s: %s',
 				$stackPtr,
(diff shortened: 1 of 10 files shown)


moorscode:
Does it always have to be an array or can you still provide a string if you only want to provide one directory?

jrfnl:
> Does it always have to be an array or can you still provide a string if you only want to provide one directory?

The short answer is: yes, you _should_ always provide an array, however, let me explain how it now works:
1. The sniff (and ruleset until upstream issue is fixed) already contains the default `doubles_path`, which is set to `/tests/doubles`.
    If a plugin just uses the default directory, nothing needs to be done and the sniff does not need to be mentioned in the custom ruleset.
2. If a plugin wants to _add_ a second `doubles_path`, while keeping the default value as well, as of PHPCS 3.4.0, this can be done using:
    ```xml
	<rule ref="Yoast.Files.TestDoubles">
		<properties>
			<property name="doubles_path" type="array" extend="true">
				<element value="/tests/src/doubles"/>
			</property>
		</properties>
	</rule>
    ```
3. If a plugin wants to *overrule* the existing `doubles_path` with a single value, this can be done using:
    ```xml
	<!-- New style array format -->
	<rule ref="Yoast.Files.TestDoubles">
		<properties>
			<property name="doubles_path" type="array">
				<element value="/tests/assets"/>
			</property>
		</properties>
	</rule>

	<!-- Old style array format -->
	<rule ref="Yoast.Files.TestDoubles">
		<properties>
			<property name="doubles_path" type="array" value="/tests/assets"/>
		</properties>
	</rule>
    ```
4. If a plugin would inadvertently use the now deprecated string version of the property (see below), the sniff has a fall-back build in to catch that and still work properly anyway.
    See: https://github.com/Yoast/yoastcs/pull/108/files#diff-b7fe49390dc2b47fff74fdfb42b26518R116
    This fall-back should be removed in YoastCS 2.0, but has been included in this PR to prevent this change being a BC-break.
    ```xml
	<rule ref="Yoast.Files.TestDoubles">
		<properties>
			<property name="doubles_path" value="/tests/assets"/>
		</properties>
	</rule>
    ```

Does that answer your question ?

PR #111, opened 2018-12-19
New `Yoast.Commenting.FileComment` sniff

jrfnl:
This sniff is a wrapper around the `Squiz.Commenting.FileComment` sniff as included in the `WordPress-Docs` ruleset.

The wrapper checks whether the file is namespaced or not.

* If the file is namespaced (whether scoped or unscoped) and has a file comment, throw a warning about the file comment being unnecessary.
* If the file is namespaced, but without a namespace name, the normal rules from the `Squiz` sniff apply.

Includes unit test testing the logic in the wrapper. The logic in the Squiz sniff is tested upstream.

Fixes #106

### Remove redundant file comments

Remove file comments from the existing PHP files in this library to comply with the new `Yoast.Commenting.FileComment` sniff.

PR #112, opened 2018-12-19
New `Yoast.Namespaces.NamespaceDeclaration` sniff

### LINKED ISSUE #110: Create sniff to check against scoped namespaces and no name namespace declarations
See for details: https://github.com/Yoast/yoastcs/issues/106

jrfnl:
This sniff:
- Forbids the use of namespace declarations without a namespace name.
- Forbids the use of scoped namespace declarations.
- Forbids having more than one namespace declaration in a file.

Includes unit tests.

Fixes #110

---

Note: the build for this sniff won't pass until PR #111 has been merged as the code in this sniff already complies with the sniff being introduced in #111.

### DIFF
--- a/Yoast/Sniffs/Namespaces/NamespaceDeclarationSniff.php
+++ b/Yoast/Sniffs/Namespaces/NamespaceDeclarationSniff.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace YoastCS\Yoast\Sniffs\Namespaces;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Verify namespace declarations.
+ *
+ * This sniff:
+ * - Forbids the use of namespace declarations without a namespace name.
+ * - Forbids the use of scoped namespace declarations.
+ * - Forbids having more than one namespace declaration in a file.
+ *
+ * {@internal This sniff might be a good candidate for pulling upstream in PHPCS
+ * itself. An issue to that effect should be opened to see if there is interest.}}
+ *
+ * @package Yoast\YoastCS
+ * @author  Juliette Reinders Folmer
+ *
+ * @since   1.2.0
+ */
+class NamespaceDeclarationSniff implements Sniff {
+
+	/**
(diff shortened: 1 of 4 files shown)


Yoast/Sniffs/Namespaces/NamespaceDeclarationSniff.php
```diff
@@ -0,0 +1,111 @@
...
+				$phpcsFile->addError(
+					'Namespace declarations without a namespace name are not allowed.',
+					$stackPtr,
+					'NoNameForbidden'
+				);
+			}
+		}
+
+		$count = count( $statements );
+		if ( count( $statements ) > 1 ) {
```
moorscode: Optimization, using `$count` instead of counting again.

Yoast/Tests/Namespaces/NamespaceDeclarationUnitTest.2.inc
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace; // Error - no name.
+
+namespace {} // Error x 3 - no name + scoped namespace declaration + not the only namespace declaration in the file.
```
moorscode, line 5: Would these cases be relevant?

```php
namespace /* some comment */ {}
```

or
```php
namespace /* some comment */;
```

or 
```php
/* some comment */ namespace;
```

Yoast/Sniffs/Namespaces/NamespaceDeclarationSniff.php
jrfnl: *grin* and that's why I created the variable, just forgot to change the second use instance.

Yoast/Tests/Namespaces/NamespaceDeclarationUnitTest.2.inc
jrfnl: The first two: yes. The sniff already takes this into account, but never a bad thing to document this in the unit tests. :+1: 

The third: no. The sniff only looks at what's after the `namespace` keyword, not what's before, so the third wouldn't be testing anything.

I'll add those two.

jrfnl:
I've fixed up the original commit to include the extra unit tests and use the `$count` variable.

moorscode:
👍 code looks good, tests look good

PR #114, opened 2018-12-22
PHPUnit: add Composer script to run the tests

jrfnl:
.. and various other tweaks.

### Unit testing the sniffs

A `dev` requirement of `phpunit/phpunit` has been added to the `composer.json` file, as well as a script to run the unit tests.
While PHPUnit is not strictly speaking a dependency of YoastCS (but of PHPCS), this just makes life easier on YoastCS developers who use the Composer install for development.

If - as a YoastCS sniff developer - you've installed YoastCS using Composer, you can now run the unit tests, like so:
```bash
composer test
```

**Note**: the `travis.yml` script does not use this script as it uses the Travis image native PHPUnit version instead of a (potentially) Composer installed one. For most builds PHPUnit wouldn't be Composer installed anyway as - except for one - all builds use `--no-dev`.

### Use the PHP version used by Composer for Composer scripts

Using the `@php` prefix in the scripts ensures that the scripts are run against the same PHP version as with which Composer is called, instead of against the system default PHP version.

This is important if - as a YoastCS dev - you change the PHP version for Composer to be able to test things against different PHP versions.

Ref: composer/composer#7645

### Minor cleanup of the `PHPCS_IGNORE_TESTS` directive

WPCS 2.0.0 will remove the `WordPress-VIP` standard. As that standard, as well as the other sub-standards of WPCS don't have their own unit tests anyway, we may as well remove references to them from the `phpunit.xml.dist` `PHPCS_IGNORE_TESTS` environment variable.

## Testing this PR;

It is recommended to test the Composer scripts as added/adjusted by this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ before_install:
     - |
       if [[ "$SNIFF" == "1" ]]; then
           composer install --dev --no-suggest
-          # The post-install-cmd script takes care of the installed_paths.
+          # The DealerDirect Composer plugin script takes care of the installed_paths.
       else
           # For testing the YoastCS native sniffs, the rest of the packages aren't needed.
           composer remove wp-coding-standards/wpcs phpcompatibility/phpcompatibility-wp phpmd/phpmd --no-update
--- a/composer.json
+++ b/composer.json
@@ -29,17 +29,19 @@
 	},
 	"require-dev": {
 		"phpcompatibility/php-compatibility": "^9.0.0",
-		"roave/security-advisories": "dev-master"
+		"roave/security-advisories": "dev-master",
+		"phpunit/phpunit": "^4.0 || ^5.0 || ^6.0 || ^7.0"
 	},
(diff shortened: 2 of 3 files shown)


jrfnl:
FYI: I've changed the name of the new Composer `script` from `run-tests` to `test` to be in line with scripts available in other Yoast plugins to run the unit tests.

andizer:
The composer.lock file is missing?

jrfnl:
@andizer Nope, it's not there on purpose. Committing the `composer.lock` file for libraries, i.e. code which is only used as a dependency, is not needed and should be discouraged. For that reason, there is no `composer.lock` file in this repo.

andizer:
Okay, I had to do a `composer update` to make phpunit work.

jrfnl:
@andizer Yes, that's correct, you would have had to anyway as the `composer.json` file has changed.

abotteram:
Acceptance 👍

PR #116, opened 2018-12-23
PHPCS/PHPCompatibility: only scan PHP files

jrfnl:
The YoastCS ruleset can be used to scan PHP, JS and CSS files, just like PHPCS and WPCS.

The PHPCompatibility ruleset, however, is only aimed at PHP files.

This small tweak makes sure that if a project contains JS and CSS files and the project based ruleset allows for those to be scanned, that the PHPCompatibility checks are only applied to PHP files.

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -77,7 +77,9 @@
 	#############################################################################
 	-->
 	<config name="testVersion" value="5.2-"/>
-	<rule ref="PHPCompatibilityWP"/>
+	<rule ref="PHPCompatibilityWP">
+		<include-pattern>*\.php$</include-pattern>
+	</rule>
 
 
 	<!--


Dieterrr:
Was CRing this and the most relevant information I could find was this: 
https://github.com/PHPCompatibility/PHPCompatibilityWP/issues/5#issuecomment-442826729 😄 

CR done 👍 

Not sure how to test this?

jrfnl:
@Dieterrr Sorry, I should have added some links, but yes, you did find a relevant issue ;-)

The `<include-pattern>` directive was introduced in PHPCS 3.0 and - similar to the `<exclude-pattern>` directive, it allows for "special" regular expressions, i.e. PCRE regex with any `*` automatically replaced with `.*`.

Relevant links in the PHPCS docs:
* https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-Ruleset (near the bottom of the big code sample)
* https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders - about the special regex syntax.

There is no "easy" way to definitively test this and you would be testing that PHPCS works correctly rather than this specific change.

PR #117, opened 2018-12-29
Use PHP_CodeSniffer v 3.4.0

jrfnl:
This updates YoastCS to use PHPCS 3.4.0 which was released this week.
See: https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.4.0

Aside from it being a new release with lots of bug fixes, there are also a few specific changes in PHPCS 3.4.0 which benefit YoastCS, most notably some new (or improved) sniffs which won't be added to WPCS yet until WPCS changes its minimum PHPCS requirement to PHPCS 3.4.0, which is not expected to happen for at least another three or four months.

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -86,10 +86,10 @@
 	These may make it into WPCS at some point. If so, they can be removed here.
 	#############################################################################
 	-->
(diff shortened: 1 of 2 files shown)


andizer:
Looks good to me!

PR #125, opened 2019-03-28
Add new Yoast.Yoast.AlternativeFunctions sniff

### LINKED ISSUE #124: Sniff to enforce usage WPSEO_Utils:format_json_encode
In https://github.com/Yoast/wordpress-seo/pull/12537, we're adding a util `WPSEO_Utils:format_json_encode()` which uses `wp_json_encode()` under the hood, adds a `JSON_UNESCAPED_SLASHES` flag if the PHP version > 5.4 and pretty prints if `WP_DEBUG` is enabled.

I want to enforce the util to be used everywhere instead of `wp_json_encode` and `json_encode`. We can also autofix.

jrfnl:
This introduces a new sniff to YoastCS: `Yoast.Yoast.AlternativeFunctions` which will initially sniff for function calls to `json_encode()` and `wp_json_encode()` and throw an `error` when these are detected, suggesting to use `WPSEO_Utils::format_json_encode()`  instead.

The sniff has been set up to easily allow for additional function groups to be added where the use of PHP/WP native functions should be discouraged in favour of using Yoast native functions.

Includes auto-fixer.
Includes unit tests.

Includes disabling the WPCS native check for `json_encode()` which recommends using `wp_json_encode()`.

Additional notes:
* The sniff will not trigger on calls to non-global functions of the same name.
* As it currently is, the sniff will throw false positives for namespaced functions imported via a `use` statement, like in the below code sample, but this is not an issue for the current Yoast code bases.
    ```php
    use function My\Namespace\some_function as wp_json_encode;

    wp_json_encode( $data );
    ```
* As it currently is, the sniff will throw a false positive for code using the `namespace` keyword as an operator.
     This will be fixed in WPCS 2.1.0 and the unit test can be uncommented once WPCS 2.1.0 has been released and YoastCS ups its minimum WPCS requirement.
* The auto-fixer will only be allowed to run when the function call detected only passes one parameter.
    Both the `json_encode()`, as well as the `wp_json_encode()` functions allow for passing three parameters `$data, $options, $depth`.
    The new Yoast native `WPSEO_Utils:format_json_encode()` method only allows for the first parameter and auto-fixing would break any function call passing more than one parameter.
    An (non-fixable) error with a separate error code will still be thrown in that case.
* The auto-fixer takes into account that code may be namespaced and will prefix the replacement with a `\` to indicate that the class name is fully qualified if it is detected that the file is namespaced.
    It will *not* add a `use` statement for the class, that is outside the scope of this sniff.
* While the duplication within the name may appear a little redundant, I've created this to mirror WPCS, where the `WP` category has sniffs regarding WP features being sniffed for, just like this sniff is sniffing for a `Yoast` specific feature (not) being used.
    Alternative category name suggestions welcome.
* Once the sniff has been merged and WPSEO has adopted the latest version of YoastCS, the call to `wp_json_encode()` from within the `WPSEO_Utils::format_json_encode()` function will need to be whitelisted for this sniff.

Fixes #124

### DIFF
--- a/Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
+++ b/Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
@@ -50,11 +50,11 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 			$replacement = $this->groups[ $group_name ]['replacement'];
 		}
 
-		$fixable = true;
-		$error   = $this->groups[ $group_name ]['message'];
-		$type    = ( $this->groups[ $group_name ]['type'] === 'error' );
-		$code    = $this->string_to_errorcode( $group_name . '_' . $matched_content );
-		$data    = array(
+		$fixable    = true;
+		$message    = $this->groups[ $group_name ]['message'];
+		$is_error   = ( $this->groups[ $group_name ]['type'] === 'error' );
+		$error_code = $this->string_to_errorcode( $group_name . '_' . $matched_content );
+		$data       = array(
 			$matched_content,
 			$replacement,
 		);
@@ -70,19 +70,19 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 				 * when only the first parameter is passed.
 				 */
 				if ( $this->get_function_call_parameter_count( $stackPtr ) !== 1 ) {
-					$fixable = false;
-					$code   .= 'WithAdditionalParams';
+					$fixable     = false;
+					$error_code .= 'WithAdditionalParams';
 				}
 
 				break;
 		}
 
 		if ( $fixable === false ) {
-			$this->addMessage( $error, $stackPtr, $type, $code, $data );
+			$this->addMessage( $message, $stackPtr, $is_error, $error_code, $data );
 			return;
 		}
 
-		$fix = $this->addFixableMessage( $error, $stackPtr, $type, $code, $data );
+		$fix = $this->addFixableMessage( $message, $stackPtr, $is_error, $error_code, $data );
 		if ( $fix === true ) {
 			$namespaced = $this->determine_namespace( $stackPtr );
 
--- a/Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
+++ b/Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace YoastCS\Yoast\Sniffs\Yoast;
+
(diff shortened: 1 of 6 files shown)


jrfnl:
I've added a second commit which fixes the build (at least it should). The new sniff uses one of the WPCS abstract classes, so WPCS has now become a dependency we actually need for the unit tests to be able to run.

Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
```diff
@@ -0,0 +1,97 @@
...
+	public function process_matched_token( $stackPtr, $group_name, $matched_content ) {
+
+		$replacement = '';
+		if ( isset( $this->groups[ $group_name ]['replacement'] ) ) {
+			$replacement = $this->groups[ $group_name ]['replacement'];
+		}
+
+		$fixable = true;
+		$error   = $this->groups[ $group_name ]['message'];
+		$type    = ( $this->groups[ $group_name ]['type'] === 'error' );
```
moorscode: This variable name suggests it contains a "type" description or object, instead it is a boolean. Suggest to rename to reflect this more: `$is_error`

Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public function process_matched_token( $stackPtr, $group_name, $matched_content ) {
+
+		$replacement = '';
+		if ( isset( $this->groups[ $group_name ]['replacement'] ) ) {
+			$replacement = $this->groups[ $group_name ]['replacement'];
+		}
+
+		$fixable = true;
+		$error   = $this->groups[ $group_name ]['message'];
```
moorscode: Any reason why this variable is not called `$message`?
As the definition can distinguish between errors and warnings, the message is not always an "error" message.

Yoast/Sniffs/Yoast/AlternativeFunctionsSniff.php
```diff
@@ -0,0 +1,97 @@
...
+		$type    = ( $this->groups[ $group_name ]['type'] === 'error' );
+		$code    = $this->string_to_errorcode( $group_name . '_' . $matched_content );
```
moorscode: Any objections for renaming this variable to `errorcode`?

moorscode (dismissed):
Couple of naming improvements/questions.
Functionally and test-wise it looks good.

jrfnl:
FYI: The WPSEO PR which inspired this sniff - https://github.com/Yoast/wordpress-seo/pull/12537 - has been merged.

moorscode:
Looks good to me 👍

PR #127, opened 2019-04-24
Remove PHP Mess Detector

jrfnl:
PHPMD is currently not used by any of the Yoast plugins, nor is the package at this time being actively maintained.

With this in mind, PHPMD will be removed from YoastCS.
It can always be brought back in the future if needed.

This PR:
* Removes the dependency from the `composer.json` file.
* Removes the PHPMD example configuration file.
* Removes any and all references to PHPMD in other places, except for the changelog.


atimmer (approved):

jrfnl:
On second check, it turns out that there are three (four) plugins which do actually use PHPMD, though not necessarily via the Composer install. These plugins provide the `phpmd.xml` file as part of the CodeClimate configuration:
* WPSEO Free (and by extension Premium)
* Local
* Comment Hacks

PR #128, opened 2019-05-08
YoastCS: add new `Yoast.Commenting.CoversTag` sniff

### LINKED ISSUE #123: Check for correct usage of @covers tags
The `@covers` tag in docblocks is use to denote what method/class a unit test tests and to prevent "incidental" coverage from being recorded.

"Incidental coverage" in this context, means tests calling methods other than the method under test to set up/prepare test. The resulting "lines covered" from this test set up/preparation should not be recorded.

Refs:
* https://phpunit.readthedocs.io/en/8.0/code-coverage-analysis.html 
* https://phpunit.readthedocs.io/en/8.0/annotations.html#covers

These tags are only effective when used correctly and various versions of PHPUnit may have varying degrees of tolerance for incorrectly used `@covers` tags.

If a `@covers` tag is ignored because it is incorrect, code coverage for incidental coverage could be recorded, skewing the code coverage figures.

I would like to propose a new sniff to check the following:
* That all unit tests have a `@covers` tag either at the class or the method level.
    A `warning` should be thrown if the `@covers` tag is missing.
* That the content of the `@covers` tag is in a valid format as specified by the PHPUnit documentation.
    An `error` should be thrown when an invalid format is encountered and whenever possible this format should be auto-fixed.

Implementation notes:
* What classes/methods should be considered test classes/methods can be derived from the PHPUnit config file, if available.
    If not available, the sniff should either bow out or use sensible defaults which can be adjusted via the PHPCS config file.
* If a test method/class does not have a docblock, bow out. That is covered by another sniff.

jrfnl:
This new sniff checks the following:
* That the contents of PHPUnit `@covers` contains a supported annotation.
   Includes a fixer for some common errors.
* That union/intersect annotations aren't used.
    If found, the annotation will be split in separate `@covers` tags.
* That a docblock doesn't contain both a `@covers` tag as well as a `@coversNothing` tag.
* That a docblock doesn't contain multiple `@coversNothing` tags.
    If found and none or only one of the annotations contains a comment, the superfluous tags will be removed.
* That a docblock doesn't contain duplicate `@covers` tags.
    If found, the superfluous tags will be removed.

Includes unit tests.

Note: a future enhancement to this sniff could be to add support for detecting class-level `@coversDefaultClass` annotations and verifying that the default class name is not used in method-level `@covers` tags, including auto-fixing those.

Partially fixes #123

### Testing

To see the effect of this new sniff:
* Check out this branch.
* Run `composer install`.
* From the root of this repo, run `vendor/bin/phpcs ./yoast/tests/commenting/coverstagunittest.inc --standard=yoast --sniffs=yoast.commenting.coverstag` to see the errors reported based on the unit tests.
* To see the fixer in action, run `vendor/bin/phpcbf ./yoast/tests/commenting/coverstagunittest.inc --standard=yoast --sniffs=yoast.commenting.coverstag`
* If you get an error try deleting the `vendor` folder and the `composer.lock` file and run `composer install` again.

Note: I have tested this sniff on a number of the plugin repos and found no issues with the sniff (and plenty in the plugins).

### DIFF
--- a/Yoast/Sniffs/Commenting/CoversTagSniff.php
+++ b/Yoast/Sniffs/Commenting/CoversTagSniff.php
@@ -1,0 +1,355 @@
+<?php
+
+namespace YoastCS\Yoast\Sniffs\Commenting;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+
+/**
+ * Verifies that a @covers tag annotation follows a format supported by PHPUnit.
+ *
+ * Also ensures that:
+ * - each @covers tag has an annotation;
+ * - there are no duplicate @covers tags;
+ * - there are no duplicate @coversNothing tags;
+ * - a method does not have both a @covers as well as a @coversNothing tag.
+ *
+ * @package Yoast\YoastCS
+ * @author  Juliette Reinders Folmer
+ *
+ * @since   1.3.0
+ */
+class CoversTagSniff implements Sniff {
+
+	/**
+	 * Regex to check for valid content of a @covers tags.
+	 *
+	 * Takes the WP naming conventions into account (up to a point).
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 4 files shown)


abotteram:
CR 👍 
Acceptance 👍

PR #130, opened 2019-05-17
PHPCS: add common exclude patterns to the ruleset

jrfnl:
As these exclude patterns are a special kind of regex which will be applied to the _absolute_ path of the file being scanned, setting these in the `YoastCS` ruleset instead of in the repo specific custom configs will work fine and will allow us to simplify the repo specific rulesets a little.

I've tested this change extensively with repo specific configs adjusted for this change and have found no problems.


herregroen:
CR 👍

PR #132, opened 2019-05-17
PHPCS: upgrade variable alignment issues from warning to error

jrfnl:
Variable alignment is a simple whitespace issue which should be easy to solve in all cases.

However, as the sniffs throw warnings instead of errors and warnings are ignored by some of the repos, these are typically issues which creep back into the repo quickly if builds don't break on them, necessitating a regular clean up again.

By changing the message type for these sniffs from `warning` to `error`, we can safeguard that these issues will be kept at bay.


IreneStr:
I agree
Acceptance 👍

PR #133, opened 2019-05-17
YoastCS/FileName: always remove the longest matching prefix when prefixes overlap

jrfnl:
The sniff allows for passing a set of `prefixes` which will be removed from an OO name to get to the expected file name.

As it was, when two `prefixes` passed would overlap - think `yoast` and `yoast_news` -, the order in which these were provided in the configuration file would determine the order in which the sniff would attempt to match these.

I.e. if the prefixes were provided in the above order and the class name was `Yoast_News_Something`, the sniff would suggest `news-something.php` as the filename, not `something.php`.

The change in this commit, changes that behaviour by ordering the `prefixes` array before using it which ensures that the _longest_ matching prefix will be used when two prefixes overlap.

Includes unit test.

Unfortunately, as the tests are more similar to integration tests rather than unit tests, the actual message is not tested, even though that's where the change lies.

### To manually test this PR and see the difference:
* Check out the `develop` branch.
* Cherrypick the new test case file.
* Temporarily add a default value to the `$prefixes` property in the sniff file:
    ```php
	public $prefixes = array(
	    'yoast',
	    'wpseo',
	    'yoast_plugin',
	);
    ```
* Make sure you have run `composer install` and then run the following command from the repo root:
    `vendor/bin/phpcs ./yoast/tests/files/filenameunittests/classes/yoast-plugin-some-class.inc --standard=Yoast --sniffs=yoast.files.filename`
* The resulting error message will read:
    ` Class file names should be based on the class name without the plugin prefix. Expected plugin-some-class.inc, but found yoast-plugin-some-class.inc.`
* Now do a hard reset on the develop branch.
* Check out this branch.
* Make the same temporary change to the default value of the `$prefixes` property as before.
* Re-run the command.
* The resulting error message should now read:
    ` Class file names should be based on the class name without the plugin prefix. Expected some-class.inc, but found yoast-plugin-some-class.inc.`

----

PR includes removing some technical debt:

### YoastCS/FileName: remove support for passing properties as strings

Both the public `exclude`, as well as the `prefixes` property for this sniff are by nature array properties.

Setting array properties within unit tests was not yet properly supported by PHPCS until version 3.3.0. For that reason, the unit tests for this sniff still used the old-style property setting annotation and the sniff itself contained code to work around the issue.

As this problem has been solved now in PHPCS, the work-arounds in the sniff can be removed and the new-style property setting can now be used in the unit tests.

### DIFF
--- a/Yoast/Tests/Files/FileNameUnitTests/ExcludedFile.inc
+++ b/Yoast/Tests/Files/FileNameUnitTests/ExcludedFile.inc
@@ -1,5 +1,6 @@
-@codingStandardsChangeSetting Yoast.Files.FileName exclude ExcludedFile.inc
+<!-- Annotation must be on line 2 as this sniff throws issues on line 1 and PHPCS ignores errors on annotation lines. -->
+phpcs:set Yoast.Files.FileName exclude[] ExcludedFile.inc
 
 <?php
 
-// @codingStandardsChangeSetting Yoast.Files.FileName exclude false
+// phpcs:set Yoast.Files.FileName exclude[]
--- a/Yoast/Sniffs/Files/FileNameSniff.php
+++ b/Yoast/Sniffs/Files/FileNameSniff.php
@@ -32,6 +32,11 @@ class FileNameSniff implements Sniff {
 	 *
 	 * These prefixes do not need to be reflected in the file name.
 	 *
+	 * Note:
+	 * - Prefixes are matched in a case-insensitive manner.
+	 * - When several overlapping prefixes match, the longest matching prefix
+	 *   will be removed.
+	 *
 	 * @var string[]
 	 */
 	public $prefixes = array();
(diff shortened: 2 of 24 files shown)


Yoast/Tests/Files/FileNameUnitTests/ExcludedFile.inc
```diff
@@ -1,5 +1,6 @@
-@codingStandardsChangeSetting Yoast.Files.FileName exclude ExcludedFile.inc
+<!-- Annotation must be on the line 2 as this sniff throws issues on line 1 and PHPCS ignores errors on annotation lines. -->
```
IreneStr: Unneeded `the` fore `line 2` (in all files below as well)

jrfnl: Fixed in the respective commits & force pushed.

IreneStr:
CR and acceptance 👍

PR #137, opened 2019-05-29
CS: don't use yoda conditions

jrfnl:
While WPCS still demands Yoda conditions, YoastCS never has.

This fixes the few instances in the code base were Yoda conditions had creeped in as code was based off code taken from WPCS.


IreneStr:
CR 👍

PR #138, opened 2019-05-29
YoastCS/AlternativeFunctions: activate disabled test

jrfnl:
See #125

This unit tests was disabled as the parent sniff from WPCS did not handle this correctly yet in WPCS < 2.1.0.

As the minimum WPCS requirement for YoastCS has gone up to WPCS 2.1.1, this test can now be enabled.


abotteram:
CR 👍 Acceptance 👍

PR #144, opened 2019-07-21
Travis: ignore PHP deprecation notices for stable PHPCS releases

jrfnl:
The unit tests will fail when a PHP warning/notice/deprecation notice is encountered.

Notices thrown by already released PHPCS versions won't get fixed anymore (in that version), so failing the unit tests on those is moot and will skew the reliability of the Travis results.

For now, I've chosen to just ignore Deprecation notices.
At a later point, this can be re-evaluated to see if more notices should be ignored for already released upstream versions.

To see the effect of this PR, compare the output from the following builds:
* Recent build against `develop`: https://travis-ci.org/Yoast/yoastcs/jobs/561796949#L565
* The same build against this branch: https://travis-ci.org/Yoast/yoastcs/jobs/561842077#L541


IreneStr:
CR + acceptance 👍

PR #145, opened 2019-07-25
PHPCS unit tests: minor fixes to the test case files

jrfnl:
Fix unintentional parse errors and annotate intentional parse errors which are there to test edge cases.


igorschoester:
CR & Acceptance 👍

PR #147, opened 2019-07-25
Readme: update the text

jrfnl:
The text in the readme was a bit dated.

Includes adding instructions on how to view the "end-user" docs for the PHPCS sniffs from the command-line. Also see: #146

PR #139, opened 2019-05-29
Changelog for the 1.3.0 release

jrfnl:
* Preliminary release date set at Thursday June 6.
* Includes all currently merged changes.

PR #148, opened 2019-07-25
GH: add pull request templates

jrfnl:
This adds three different pull request templates:
1. A dependency change template.
2. A sniff change template.
3. A generic template for anything that doesn't fit in the above two categories.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/dependency_change.md
+++ b/.github/PULL_REQUEST_TEMPLATE/dependency_change.md
@@ -1,0 +1,19 @@
+---
+name: Dependency change
+about: Update one or more of the YoastCS dependencies
+labels: "yoast cs/qa"
+---
+
+## Description
+
+*
+
+Refs:
+<!-- Add relevant links to, for instance, the changelog of the updated dependency. -->
+*
+
+<!--
+If the dependency update involves PHP_CodeSniffer or the WordPress Coding Standards,
+please make sure that the `.travis.yml` script is also updated to reflect this change,
+so the unit tests are run against the relevant supported versions of these repos.
+-->
--- a/.github/PULL_REQUEST_TEMPLATE/generic.md
+++ b/.github/PULL_REQUEST_TEMPLATE/generic.md
@@ -1,0 +1,6 @@
+---
+name: General
+about: Change which doesn't fit any of the other categories
+labels: "yoast cs/qa"
+---
+
--- a/.github/PULL_REQUEST_TEMPLATE/sniff_change.md
+++ b/.github/PULL_REQUEST_TEMPLATE/sniff_change.md
@@ -1,0 +1,41 @@
+---
+name: Sniff change
+about: Introduce a new sniff or update an existing sniff
+labels: "yoast cs/qa"
+---
+
+## Description
+
+*
+
+
+Refs:
+<!-- Add relevant links to, for instance, related upstream issues. -->
+*
+
+
+## Test instructions
+<!--
+Please follow these guidelines when creating test instructions:
+- Please provide step-by-step instructions how to reproduce the issue, if applicable.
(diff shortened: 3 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE/dependency_change.md
```diff
@@ -0,0 +1,19 @@
...
+
+Refs:
+<!-- Add relevant links to, for instance, the changelog of the updated dependency. -->
+*
+
+<!--
+If the dependency update involves PHP_CodeSniffer or the WordPress Coding Standards,
+please make sure that the `.travis.yml` script is also updated to reflect this change,
+so the unit tests are run against the relevant supported versions of these repos.
+-->
```
IreneStr, line 19: I think test instructions could be relevant in this template as well: check whether this and this is done in the travis build etc. Having a test instructions heading in the template forces the PR writer to think about testing.

.github/PULL_REQUEST_TEMPLATE/generic.md
```diff
@@ -0,0 +1,6 @@
+---
+name: General
+about: Change which doesn't fit any of the other categories
+labels: "yoast cs/qa"
+---
+
```
IreneStr, line 6: Aren't a description, refs and test instructions always relevant for a PR? I'd expect those headings in the generic template.

.github/PULL_REQUEST_TEMPLATE/sniff_change.md
```diff
@@ -0,0 +1,41 @@
...
+## Description
+
+*
+
+
+Refs:
+<!-- Add relevant links to, for instance, related upstream issues. -->
+*
+
+Fixes #
```
IreneStr: In all other Yoast repos, this is the final line of the template. Let's go for consistency.

IreneStr (changes requested):
CR 🚧 Some comments and questions

.github/PULL_REQUEST_TEMPLATE/dependency_change.md
jrfnl: The point of the remark about updating the Travis script, is that a dependency change on this repo does not need manual testing, but will automatically be tested via the Travis builds.

From another perspective: testing these type of changes for this repo isn't necessary in the first place.

All builds are tested against high/low versions of the current dependencies, with `high` being the current `develop` versions of the dependencies.

The only dependency changes which can be made in that sense, is the _raising_ of the minimum supported version. And that new minimum supported version has basically already been tested via Travis for a while.

Last point: the only exception to this would when it wouldn't be an "update" of one of the dependencies, but replacing a dependency or adding a new one, but that is not the situation targetted by this template.

.github/PULL_REQUEST_TEMPLATE/sniff_change.md
jrfnl: Will do, though I still find that illogical in all other repos.

The issue being fixed will normally contain additional information which can enrich the PR description and inform the testing of the PR, so IMHO having it between description and test instructions seems way more in line with flow.

.github/PULL_REQUEST_TEMPLATE/generic.md
jrfnl: Not really. Take the following PRs as examples of PRs which would fit in the `generic` template and where - other than a description, but that is implied - nothing else is needed:
#127, #137, #139, #147

jrfnl: Not really. Take the following PRs as examples of PRs which would fit in the `generic` template and where - other than a description, but that is implied - nothing else is needed:
#127, #137, #139, #147

IreneStr:
CR 👍

PR #151, opened 2019-08-06
CS: use short arrays

jrfnl:
WPCS will start forbidding the use of short arrays as of WPCS 2.2.0.

For YoastCS the decision has been taken to not comply with this rule and to enforce short arrays instead.

As the minimum supported PHP version is still PHP 5.2 for most repos, we cannot yet enforce this through the `YoastCS` ruleset.

So, for now, this will only be enforced for PHP 5.4+ directories in those repos which have them.

As YoastCS itself has a minimum PHP requirement of PHP 5.4, we _can_ start enforcing it in the code of YoastCS itself.
This PR implements that.


IreneStr:
CR 👍

PR #156, opened 2019-10-31
New `ObjectNameDepth` sniff

### LINKED ISSUE #142: New ClassName sniff
As discussed today with @omarreiss @herregroen @IreneStr, a new sniff needs to be created which will check that _**the names of classes declared within a namespace**_ have a maximum of two underscores in them, i.e they can consist of maximum three words.

The sniff should contain separate properties for `recommended` (warning) and `maximum` (error) number of underscores.
The default will be `2` in both cases.
Individual plugins can use the properties to deviate.

Other than that, the normal WPCS class names rules (`Camel_Caps`, acronyms allowed) apply.

### Valid code samples
```php
class Non_Namespaced_Long_Class_Name {}
```
```php
namespace Yoast\WP\Plugin;

class Short_Class_Name {}
```

### Invalid code samples
```php
namespace Yoast\WP\Plugin;

class Much_Too_Long_Class_Name {}
```

jrfnl:
This new sniff checks that the names of declared objects - classes, interfaces, traits - contain no more than three words / two underscores.

This rule is only applied to objects declared in a namespace. For non-namespaced objects the rule does not apply.

The sniff will throw a `warning` when the name consists of more words than `recommended_max_words` and an `error` when the name is longer than `max_words`.

Both `recommended_max_words`, as well as `max_words` are configurable and can be changed from a custom ruleset.

By default, `recommended_max_words` and `max_words` are set to the same value: `3`, i.e. three words.

Note: the sniff currently extends the WPCS `Sniff` class to allow for using one of the utility methods contained in WPCS. Once `PHPCSUtils` comes out and use of `PHPCSUtils` is implemented in YoastCS, this will no longer be necessary and the (improved) utility method in `PHPCSUtils` can be used instead.

Includes unit tests.
Includes documentation.

Fixes #142

### DIFF
--- a/Yoast/Sniffs/NamingConventions/ObjectNameDepthSniff.php
+++ b/Yoast/Sniffs/NamingConventions/ObjectNameDepthSniff.php
@@ -87,7 +87,7 @@ class ObjectNameDepthSniff extends WPCS_Sniff {
 
 		$comment_end = $this->phpcsFile->findPrevious( $find, ( $stackPtr - 1 ), null, true );
 		if ( $this->tokens[ $comment_end ]['code'] === T_DOC_COMMENT_CLOSE_TAG ) {
-			// Only check is the class has a docblock.
+			// Only check if the class has a docblock.
 			$comment_start = $this->tokens[ $comment_end ]['comment_opener'];
 			foreach ( $this->tokens[ $comment_start ]['comment_tags'] as $tag ) {
 				if ( $this->tokens[ $tag ]['content'] === '@deprecated' ) {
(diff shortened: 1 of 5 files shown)


Yoast/Sniffs/NamingConventions/ObjectNameDepthSniff.php
```diff
@@ -0,0 +1,129 @@
...
+		// Check if the class is deprecated.
+		$find = [
+			T_ABSTRACT   => T_ABSTRACT,
+			T_FINAL      => T_FINAL,
+			T_WHITESPACE => T_WHITESPACE,
+		];
+
+		$comment_end = $this->phpcsFile->findPrevious( $find, ( $stackPtr - 1 ), null, true );
+		if ( $this->tokens[ $comment_end ]['code'] === T_DOC_COMMENT_CLOSE_TAG ) {
+			// Only check is the class has a docblock.
```
atimmer:
```suggestion
			// Only check if the class has a docblock.
```

atimmer (approved):
Looking good!

jrfnl:
Thanks ;-)

PR #158, opened 2019-10-31
Ruleset: change minimum versions to PHP 5.6 and WP 5.2

jrfnl:
YoastCS 2.0 is intended for the Yoast plugin after the drop of support for WP < 5.2, which means that the minimum supported PHP version for the plugins will now be PHP 5.6 and the minimum supported WP version will be WP 5.2.


andizer:
👍

PR #159, opened 2019-10-31
Ruleset: demand short arrays

jrfnl:
As previously discussed and approved with @herregroen and @omarreiss

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -62,6 +62,9 @@
 		<exclude name="Yoast.Commenting.FileComment.AuthorTagOrder"/>
 		<exclude name="Yoast.Commenting.FileComment.CopyrightTagOrder"/>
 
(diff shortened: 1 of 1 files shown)


Yoast/ruleset.xml
```diff
@@ -108,6 +111,10 @@
 		</properties>
 	</rule>
 
+	<!-- In contrast to WPCS: demand short arrays. -->
+	<rule ref="Generic.Arrays.DisallowLongArraySyntax"/>
+
+
```
andizer, line 117: There is a bonus newline?

jrfnl: Sort of. In these config files, I normally have one new line between rules and two new lines before the start of a new section.

The extra new line fixes a minor inconsistency regarding that for this file.

andizer: I see 👍

PR #160, opened 2019-10-31
New `ValidHookName` sniff

### LINKED ISSUE #143: New HookName sniff
As discussed today with @herregroen @IreneStr and with some changes after further discussion with @omarreiss:

* Hook names in the Yoast plugin repos should start with the plugin prefix in namespace format, followed by the hook name in lowercase with words separated by underscores.
    I.e. `Yoast\WP\Free\hook_name`, not `yoast_free_hook_name`.
    To that end, the configuration of the WPCS `ValidHookName` sniff in YoastCS should be adjusted to allow for the `\` character as a word separator as well as for underscores.
* The `hook_name` part should be descriptive for the (dev-)user and does not need to follow the namespace or file path of the file they are in.
* The `hook_name` part should have a maximum of three underscores in them, i.e they can consist of maximum four words.

The sniff should contain separate properties for `recommended` (warning) and `maximum` (error) number of underscores.
The default will be `3` in both cases.
Individual plugins can use the properties to deviate.

### Valid code sample
```php
do_action( 'Yoast\WP\Plugin\hook_name' );
```

### Invalid code samples
```php
do_action( 'yoast_plugin_hook_name' );
do_action( 'Yoast\WP\Plugin\hook_name_which_is_too_long' );
```

jrfnl:
This new sniff extends the upstream WPCS `ValidHookName` sniff and overloads the `transform()` method to allow for the new Yoast specific hook prefix requirements as per issue #143.

Hooks in the Yoast plugins should still comply with the "_hook names should be lowercase with words separated by underscores_" rule, but the `Yoast\WP\PluginName` prefix should be exempt.

The overload of the `transform()` method prevents the upstream sniff from throwing errors for the prefix, while still allowing it to throw errors for the actual hook name.

Also, and in contrast to the WPCS native sniff, when the YoastCS version of this sniff has received valid prefix(es), the hook name checks will only be executed for hooks which have a valid prefix.
Non-prefixed hooks or hooks with another prefix will be ignored to prevent errors being thrown about hook names which are outside of our control.

The sniff uses the `CustomPrefixesTrait` to allow setting the `public` `$prefixes` property and to validate the input received for this property.

In addition to this, the new/extended sniff adds a couple of new checks:
- A check that the "namespace"-style prefix is used.
    The sniff allows for passing both "namespace"-like, as well as "old-style" prefixes to allow for a transition period.
    However, if an "old-style" prefix is used, rather than the "namespace"-like prefix, a `warning` will be thrown.
- A check to verify that hook names contain no more than four words / three underscores, after the plugin specific prefix.
    - The sniff will throw a `warning` when the name consists of more words than `soft_maximum_depth` and an `error` when the name is longer than `maximum_depth`.
    - Both `soft_maximum_depth`, as well as `maximum_depth` are configurable and can be changed from a custom ruleset.
        By default, `soft_maximum_depth` and `maximum_depth` are set to the same value: `4`, i.e. four words.

Note:
For dynamic hook names where the hook name length can not reliably be determined, the sniff will throw a `warning` at severity `3` suggesting the hook name be inspected manually.

As the default severity for PHPCS is `5`, this `warning` at severity `3` will normally not be shown. It is there to allow for intermittently checking of the dynamic hook names. To trigger it, `--severity=3` should be passed on the command line.

Also note:
Just like the lowercase/underscore hook name checks, the hook name length check will only be executed when the hook name starts with (one of) the plugin specific prefix.

Includes unit tests.
Includes documentation.

Fixes #143

----

This PR includes adding a `CustomPrefixes` trait.

This trait can be used by sniffs which need access to the custom plugin prefixes.

The trait contains the `public` property to allow for setting the `$prefixes` from within a custom ruleset, as well as a couple of utility methods to validate and clean the received `$prefixes` so they are usable by the sniffs.

The trait is tested via the sniffs using it.

Both this sniff as well as the upcoming `NamespaceName` sniff will use this trait.

### DIFF
--- a/Yoast/Tests/NamingConventions/ValidHookNameUnitTest.inc
+++ b/Yoast/Tests/NamingConventions/ValidHookNameUnitTest.inc
@@ -1,0 +1,124 @@
+<?php
+
+/*
+ * Verify that the original functionality of the upstream sniff is still working as expected.
+ * Functionality when no prefixes have been passed, should be unchanged.
+ */
+
+do_action( $Hook_name ); // OK.
+do_action( "admin_head" ); // OK.
+do_action( 'admin_head_' .  get_ID() . '_action' ); // OK.
+do_action( "admin_head_$Post" ); // OK.
+do_action( 'prefix_block_' . $block['blockname'] ); // OK.
+
+do_action( "adminHead" ); // Error - use lowercase.
+do_action( 'admin_Head_' . $Type . '_Action' ); // Error - use lowercase.
+do_action( "admin_head-$hook_suffix" ); // Warning - use underscore.
+apply_filters( 'adminHead', $variable ); // Error - use lowercase.
+do_action( "admin_Head_{$post->ID}_Action" ); // Error - use lowercase.
+do_action( "admin_Head_{$obj->Values[3]->name}-Action_{$obj->Values[3]->name}_Action" ); // Error - use lowercase + warning about dash.
+
+/*
+ * Verify the Yoast specific changes to allow for the plugin prefix, while still checking
+ * that the rest of the hook name is lowercase and underscore separated.
+ *
+ * @phpcs:disable Yoast.NamingConventions.ValidHookName.WrongPrefix
+ */
+
+// phpcs:set Yoast.NamingConventions.ValidHookName prefixes[] Yoast\WP\Plugin,yoast_plugin
+
+apply_filters_ref_array( 'yoast_plugin_some_hook_name', $var ); // OK.
+apply_filters( 'Yoast\WP\Plugin\some_hook_name', $var ); // OK.
+
+do_action_ref_array( 'Yoast\WP\Plugin\some-hook\name', $var ); // Warning - disallowed word separators (after prefix).
+apply_filters( 'Yoast\WP\Plugin\some_HookName', $var ); // Error - uppercase chars (after prefix).
+apply_filters_ref_array( 'yoast_plugin_Some-Hook-Name', $var ); // Error + warning, uppercase chars + dashes (after prefix).
+
+// OK, ignored. Hooks do not start with correct prefix, while valid prefixes have been passed.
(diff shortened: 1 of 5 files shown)


Yoast/Tests/NamingConventions/ValidHookNameUnitTest.inc
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * These checks are only in effect if a prefix is found.
+ * The WPCS PrefixAllGlobals sniff checks that a prefix is used, that's outside the scope of this sniff.
+ */
+
+/*
+ * Simple strings, no prefix. Includes testing handling of comments.
+ */
+do_action(
+	// phpcs:ignore Stnd.Cat.Sniff -- For reasons.
```
atimmer, line 63: What is the purpose of this line?

atimmer (approved):
Nice, looking good to merge. I have one question that doesn't block merging, but can you still answer that?

Yoast/Tests/NamingConventions/ValidHookNameUnitTest.inc
jrfnl: This is the test case file for the sniff.

A PHPCS test case file should contain good and bad code to test the sniff.
Also: a sniff cannot presume that other rules within the same ruleset will be active, so should contain as much convoluted code as is necessary to test the sniff properly.

With this particular test case, I'm testing that the sniff will still recognize the hook name correctly, even when there are comments and PHPCS annotations within the same parameter.

PHPCS annotations are tokenized differently from comments, which is why I have both in the test case code.

As I don't want this sniff to actually ignore anything, but just want to test that the sniff handles itself correctly when any PHPCS annotation is found in the parameter, I'm ignoring a non-existent sniff, just to have a PHPCS annotation token in the test case.

Does that explain it well enough ?

atimmer: Exhaustively 😁👍

PR #162, opened 2019-10-31
Travis: don't allow builds against PHP 7.4 to fail

jrfnl:
While PHPCS itself is only fully compatible with PHP 7.4 as of PHPCS 3.5.0, the issues in PHPCS don't affect the YoastCS code, so YoastCS builds against PHP 7.4 should not be allowed to fail anymore now PHP 7.4 is about to come out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,10 +78,6 @@ matrix:
     - php: 5.4
       env: PHPCS_BRANCH="3.3.1" WPCS="dev-develop"
 
-  allow_failures:
-    # Allow failures for unstable builds.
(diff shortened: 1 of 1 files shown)


andizer:
No idea how to test this, but it seems to be okay 👍

jrfnl:
> No idea how to test this, but it seems to be okay +1

@andizer Not much _to_ test. Checking the Travis output to confirm that PHP 7.4 is no longer marked as _allowed to fail_ and that the build for PHP 7.4 passes (detailed Travis script output check) was all that could be done really.

PR #163, opened 2019-11-01
New `NamespaceName` sniff

### LINKED ISSUE #141: New Namespace name sniff
As discussed today with @omarreiss @herregroen @IreneStr, a new sniff needs to be created which will check that namespace declarations comply with the following rules:

A namespace should be build up of:
1. A plugin specific prefix. This should be the namespace prefix as already specified for the WPCS `PrefixAllGlobals` sniff, like for instance `Yoast\WP\Free`.
2. The directory path to the file.
    If the plugin root is not the starting point for this, the sniff should allow for setting one or more subdirectories as starting points.
    These starting points should be provided as relative paths in relation to the project root.
    For example:
    - Standard setup: `root/admin/domain/file.php` would translate to the following namespace: `Yoast\WP\Plugin\Admin\Domain`.
    - A setup with all plugin files in a `src` directory, where the `src` directory has been indicated as a starting point:
        `root/src/admin/domain/file.php` would also translate to that same namespace: `Yoast\WP\Plugin\Admin\Domain`.
3. The namespace depth - excluding the plugin prefix - should be no more than `3` deep. If it is, an `error` will be thrown.
4. The namespace depth - excluding the plugin prefix - is recommended to be no more than `2` deep. If it is between the _recommended_ and the _maximum_ value, a `warning` will be thrown.

Both the _recommended_ as well as the _maximum_ namespace depth should be configurable properties of the sniff, so individual plugins have the ability to deviate from the "namespace depth" rules.

### Not covered by the sniff

Other than that, the namespace declaration should comply with the following additional rules which will not be covered by this sniff.
1. Each leaf of a namespace name should be in `Camel_Caps`, i.e. each word should start with a capital and words should be separated by an underscore.
    Consecutive caps, like for acronyms, will be allowed.
2. There should be no whitespace or comments within the name part of the namespace declaration.
3. There should be exactly one space between the `namespace` keyword and the start of the namespace name in a namespace declaration.
4. There should be exactly one blank line before a namespace declaration.
5. There should be at least one blank line after a namespace declaration.

These rules will be proposed for WPCS shortly and are expected to be implemented in a sniff in WPCS upstream.

### Valid code example
```php
// file: plugin-root/admin/domain-scheme/file.php
<?php
/**
 * File docblock.
 */

namespace Yoast\WP\Free\Admin\Domain_Scheme;

use ...
```

### Invalid code examples
```php
// file: plugin-root/admin/domain-scheme/file.php
<?php
/**
 * File docblock.
 */
namespace NonYoastPrefix \
	/* Here comes the real namespace */ Admin\DomainScheme\SubDomain\TooLong;
use ...
```

jrfnl:
This new sniff checks that the names of declared namespaces contain no more than three levels after the plugin specific prefix.

The sniff will throw a `warning` when the name consists of more words than `recommended_max_levels` and an `error` when the name is longer than `max_levels`.

Both `recommended_max_levels`, as well as `max_levels` are configurable and can be changed from a custom ruleset.

By default, `recommended_max_levels` and `max_levels` are set to respectively 2 and 3.

For the purposes of this sniff, `Tests\` and `Tests\Doubles\` in the namespace name will be disregarded for this level depth check, but only when these are found directly after the prefix.

Additionally, the sniff checks that the levels in the namespace name reflect the directory path of the file based on the project root as identified via the `basepath` setting.

Additional `root` directories can be indicated by setting the `$src_directory` property.

The `CustomPrefixes` trait is used to allow for setting and validating the plugin specific prefix to take into account. (same as in #160)

Includes unit tests.
Includes documentation.

Includes updating the custom ruleset used to check the code in YoastCS itself.

Fixes #141


Yoast/Docs/NamingConventions/NamespaceNameStandard.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
+<documentation title="Valid Hook Name">
```
atimmer:
```suggestion
<documentation title="Namespace Name Depth">
```

Yoast/Sniffs/NamingConventions/NamespaceNameSniff.php
```diff
@@ -0,0 +1,358 @@
...
+ * This sniff checks that the levels do not exceed the recommended and maximum depth
+ * as well as that the levels directly translate to the (sub-)directory a file is
+ * placed in.
+ *
+ * @package Yoast\YoastCS
+ * @author  Juliette Reinders Folmer
+ *
+ * @since   2.0.0
+ */
+class NamespaceNameSniff  implements Sniff {
```
atimmer:
```suggestion
class NamespaceNameSniff implements Sniff {
```

atimmer (approved):
Two small changes I can change myself with GitHub. 👍

atimmer:
Rebased to keep the history tidy 💅

jrfnl:
> 
> 
> Two small changes I can change myself with GitHub. +1

@atimmer Thanks for catching those & fixing those up :+1:

PR #172, opened 2019-12-02
Ruleset: disallow "// end" comments

jrfnl:
## Proposal / Rule addition

In the olden days, before there were good IDEs, it was often customary to have `// End class`/`// End function` etc comments on the line of a class/function/control structure closing brace to make its easier to find the end of such a structure.

As modern IDEs are perfectly capable of handling this and handle this well, those type of comments are no longer necessary and can be considered unnecessary "noise".

The sniff which is proposed to be added via this PR enforces that closing braces of classes/interfaces/traits/functions are not followed by a comment or statement, in line with PSR-12.

### Impact: Low

There are some repos which have some stray end comments, but all together not that many.

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -144,6 +144,9 @@
 	<!-- CS: ensure exactly one blank line before each property declaration. -->
 	<rule ref="Squiz.WhiteSpace.MemberVarSpacing"/>
 
+	<!-- CS: don't allow "// end class" comments and the likes. -->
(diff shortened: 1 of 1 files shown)


IreneStr (approved):
No objections :) What do you think, @andizer?

andizer:
Yes, please 😄

jrfnl:
@andizer The current sniff won't catch all (yet), but it's a start. I'll see about adding a more extensive sniff to PHPCSStandards in the near future.

PR #173, opened 2019-12-02
Ruleset: check spacing around spread operator

jrfnl:
## Proposal / Rule addition

PHPCS 3.5.0 includes a new sniff to check that the spacing between the spread operator and the variable/function call it applies to is consistent.

The default behaviour of this sniff is to not allow whitespace between the spread operator and next token.

It is expected that this sniff will eventually be added to WPCS as well.

Refs:
* squizlabs/PHP_CodeSniffer#2548
* WordPress/WordPress-Coding-Standards#1762

### Impact: None

At this moment this would not cause any errors to be thrown, but will prevent inconsistency creeping in in the future.


IreneStr (approved):
Sounds good to me as well. Do you agree, @andizer?

PR #176, opened 2019-12-06
Ruleset: enforce consistent indentation of chained method calls

jrfnl:
## Proposal / Rule addition

When object method calls are chained over multiple lines, indentation should always be one more or less than the previous and always at least one in from the start of the chain.

This addition enforces that.

Example of correct chain indentation:
```php
$obj->method()
    ->foo()
        ->bar()
    ->select();
```

Example of incorrect chain indentation
```php
$obj->method()
    ->foo()
                    ->bar()
        ->select();
```

### Impact: Low

There are some repos which have some chained method calls which don't comply, but all together not that many.

### DIFF
--- a/Yoast/ruleset.xml
+++ b/Yoast/ruleset.xml
@@ -150,6 +150,13 @@
 	<!-- CS: don't allow "// end class" comments and the likes. -->
 	<rule ref="PSR12.Classes.ClosingBrace"/>
 
+	<!-- CS: enforce consistent indentation of chained object method calls. -->
+	<rule ref="PEAR.WhiteSpace.ObjectOperatorIndent">
+		<properties>
+			<property name="multilevel" value="true"/>
+		</properties>
+	</rule>
+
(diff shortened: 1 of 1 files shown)


IreneStr:
@jrfnl In the example: is `bar` a function of `foo` or is the indentation off? Other than that: approved!

jrfnl:
> 
> 
> @jrfnl In the example: is `bar` a function of `foo` or is the indentation off? Other than that: approved!

`bar()` is a function of `$obj`. See this article for some info on method chaining: https://www.unleashed-technologies.com/blog/2009/12/19/method-chaining-php

And yes, the indentation is off which is what the sniff in this PR will start sending error messages about.

The 1-in indentation is the first example is acceptable and may be done for readability in specific cases, think a situation where the `bar()` method would create a subset of a property set in `foo()`.

IreneStr:
@jrfnl Thanks the explanation. My question was about the first example. That wasn't really clear 😅. @andizer and I approve!

PR #177, opened 2019-12-06
Ruleset: disallow leading backslashes for import use statements

jrfnl:
## Proposal / Rule addition

An import `use` statement must always be fully qualified, so a leading backslash is redundant and discouraged by PHP itself.

This sniff addition enforces this and is expected to go into WPCS at some point in the future.

Correct:
```php
namespace A\B;

use B\C;
```

Incorrect:
```php
namespace A\B;

use \B\C;
```

### Impact: None

At this moment this would not cause any errors to be thrown as any previously introduced issues have already been fixed (and there have been numerous), but the sniff will prevent new issues being introduced in the future.


IreneStr:
Fine by @andizer and me 👍

PR #178, opened 2019-12-09
Ruleset: enforce PHP open tag on a line by itself

jrfnl:
## Proposal / Rule addition

Enforce PHP open tag on a line by itself in PHP-only files.

This is partially enforced already, this just safeguards it for the edge cases which weren't enforced yet.

### Impact: None

At this moment this would not cause any errors to be thrown, but the sniff will prevent new issues being introduced in the future.


IreneStr:
Approved!

PR #182, opened 2019-12-17
Update .gitattributes

jrfnl:
* GitHub templates do not need to be included in the distributed package.
* The test case files should be treated the same as other `php` files.


karlijnbok:
CR: 👍

PR #204, opened 2020-04-08
ValidHookName: add metrics for old/new style and word count

jrfnl:
Metrics are a way to get statistics about a code base and can inform the values to use in the custom properties.

The metrics can be seen by running PHPCS with the `--report=info` option.

This PR adds two metricss to the sniff:
* Whether old-style `plugin_prefix` or new-style `plugin\prefix` prefixes are used.
* Word count in the hook name _after_ the prefix.

Notes:
* The metrics only look at the hooks using one of the allowed plugin prefixes.
    Hooks using a different prefix are disregarded (as not belonging to this plugin).
* The metrics for "word count" only apply to hooks using the new-style namespace based prefix.


bintzandt:
LGTM 👍

PR #219, opened 2020-07-23
Travis: change from "trusty" to "xenial" and test against WPCS `master`

jrfnl:
### Travis: change from "trusty" to "xenial"

As the "trusty" environment is no longer officially supported by Travis, let's remove the key which will let the script use the default distro - currently `xenial`.

This updates the Travis config to:
* Use the `xenial` distro, which at this time is the default.
* Sets the distro for low PHP versions explicitly to `trusty`.
* Makes the expected OS explicit (linux).
* Updates the `matrix` key to `jobs`, which is the canonical for which `matrix` is an alias.
* Run the "sniff" stage on PHP 7.4 instead of PHP 7.3.

### Travis: test against WPCS `master` not `develop`

The WordPressCS `master` branch contains the last stable release, which is the highest YoastCS supported WPCS version.

As development on WPCS 3.0.0 has started and WPCS 3.0.0 will contain breaking changes, testing against WPCS `develop` would currently fail.

This will be addressed once WPCS 3.0.0 has stabilized and an RC has been released.

### Travis: add test builds against unstable dependencies

This adds two additional test builds:
* One against WPCS `develop`.
* One against PHP `nightly` (PHP 8).

Both are allowed to fail and are intended to provide early warning for upcoming changes we need to account for.

### Travis: ignore platform requirements for nightly

To allow the most recent supported PHPUnit version to install on PHP nightly, the `composer install` needs to run with `--ignore-platform-reqs`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,10 @@ php:
 
 env:
   # Test against the highest/lowest supported PHPCS and WPCS versions.
-  - PHPCS_BRANCH="dev-master" WPCS="dev-develop" PHPLINT=1
+  # Using WPCS 'master' for the last stable release.
+  - PHPCS_BRANCH="dev-master" WPCS="dev-master" PHPLINT=1
   - PHPCS_BRANCH="dev-master" WPCS="2.2.0"
-  - PHPCS_BRANCH="3.5.0" WPCS="dev-develop"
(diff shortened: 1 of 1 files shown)


bintzandt:
CR: :+1: 
Acc: checked the Travis build and logs, seems to be working fine. One test failed, but that was allowed (and expected since it is against the development branch).

PR #220, opened 2020-08-05
YoastCS ruleset: set the minimum supported WP version to 5.4

jrfnl:
.. in line with the change made to the signalled minimum supported WP version in the plugins.


enricobattocchi:
OK!

PR #224, opened 2020-08-28
NamespaceName: allow for non-conventional test directory setup

jrfnl:
The tests in the Yoast repos are expected to live in a `tests` directory of the project root and the namespace should reflect that , i.e. `[Prefix]\Tests\NamespaceName`.

In the YoastSEO Free repo, however, the tests are split into two additional subdirectories, `unit` and `integration` and the ruleset configuration is setup to allow for this.

This did however mean that the allowance for an extra namespace level for `Doubles` directories did not carry through.

This small fix changes that and will prevent unwarranted namespace level depth errors for classes in the `tests\unit\doubles` directory.


karlijnbok:
CR: 👍

PR #243, opened 2021-05-27
ValidHookName: improve handling of escaped slashes in prefixes

### LINKED ISSUE #242: ValidHookName: verify correct handling of single vs double quotes names
Inspired by a question in Slack from @Djennez :

The NamingConventions/ValidHookName sniff needs a quick review for correct handling of double backslashes.

When single quoted strings are used, the "namespace-like" prefix for hooks should be fine.
However, when double quoted strings are used, the namespace separator(s) in the hook name prefix need to be escaped as they could otherwise result in a PHP escape character in the hook name.

```php
// Correct (single quoted).
apply_filter( 'Namespace\Like\Prefix\hookname', $var);

// Incorrect (double quoted - the `\` needs to be escaped).
apply_filter( "Namespace\Like\Prefix\hookname", $var);

// Correct (double quoted).
apply_filter( "Namespace\\Like\\Prefix\\hookname", $var);
```

The sniff should:
1. Not throw false positives when the prefix is correct, but uses escaped backslashes in a double quoted string.
2. Possibly throw an error when a double quoted string is used for the hook name, but the backslashes aren't escaped.

Implementation note:
* This requires that the actual surrounding quotes for the text string is checked as PHPCS will tokenize double quotes strings without interpolation as single quoted strings. (`T_CONSTANT_ENCAPSED_STRING` versus `T_DOUBLE_QUOTED_STRING`).

jrfnl:
This commit is the result of a review of the sniff for correct handling of double slashes in hook names.

When the "namespace-like" prefix type is used and the hook name is in double quotes, the "namespace separator" slash should be escaped to prevent it from accidentally escaping the **next** character.

To that end:
* Slash-escaped backslashes in prefixes will now be properly recognized and handled by the sniff and will not lead to false positives due to unrecognized prefixes.
* If the hook name is passed as a double quoted string, an extra check will be executed to verify that all backslashes are slash-escaped.
    If this is not the case, a warning will be thrown.

Includes a minor efficiency fix as the prefix is now only checked and saved to the property earlier instead of the hook name being analyzed twice for this.

Includes unit tests for the functionality and some minor adjustments to the existing unit tests.

Fixes #242

In a further iteration on this sniff, the new warning could potentially be made auto-fixable.

### DIFF
--- a/Yoast/Sniffs/NamingConventions/ValidHookNameSniff.php
+++ b/Yoast/Sniffs/NamingConventions/ValidHookNameSniff.php
@@ -61,14 +61,6 @@ class ValidHookNameSniff extends WPCS_ValidHookNameSniff {
 	public $recommended_max_words = 4;
 
 	/**
-	 * Whether this is the first text string passed to the `transform()`
-	 * method, i.e. the text string which should have the prefix.
-	 *
-	 * @var bool
-	 */
-	private $remove_prefix = false;
-
-	/**
 	 * The prefix found (if any).
 	 *
 	 * @var string
@@ -82,6 +74,17 @@ class ValidHookNameSniff extends WPCS_ValidHookNameSniff {
 	 * @var bool
 	 */
 	private $first_string = '';
+
+	/**
+	 * The quote style used for the prefix part of the hook name.
+	 *
+	 * A double quoted string without variable interpolation will be tokenized as
+	 * `T_CONSTANT_ENCAPSED_STRING`, but due to the double quotes, will still need
+	 * for namespace separators to be escaped, i.e. double-slashed.
+	 *
+	 * @var string
+	 */
+	private $prefix_quote_style = '';
 
 	/**
 	 * Process the parameters of a matched function.
@@ -101,9 +104,9 @@ class ValidHookNameSniff extends WPCS_ValidHookNameSniff {
 		 * Reset the properties which help manage this each time a new function call
 		 * is encountered.
 		 */
-		$this->remove_prefix = true;
-		$this->found_prefix  = '';
-		$this->first_string  = '';
+		$this->found_prefix       = '';
+		$this->first_string       = '';
+		$this->prefix_quote_style = '';
 		$this->validate_prefixes();
 
 		/*
@@ -115,10 +118,20 @@ class ValidHookNameSniff extends WPCS_ValidHookNameSniff {
 			$found_prefix    = '';
 
 			if ( isset( Tokens::$stringTokens[ $this->tokens[ $first_non_empty ]['code'] ] ) ) {
-				$content = \trim( $this->strip_quotes( $this->tokens[ $first_non_empty ]['content'] ) );
+				$this->prefix_quote_style = $this->tokens[ $first_non_empty ]['content'][0];
+				$content                  = \trim( $this->strip_quotes( $this->tokens[ $first_non_empty ]['content'] ) );
+
 				foreach ( $this->validated_prefixes as $prefix ) {
-					if ( \strpos( $content, $prefix ) === 0 ) {
+					if ( \strpos( $prefix, '\\' ) === false
+						&& \strpos( $content, $prefix ) === 0
+					) {
 						$found_prefix = $prefix;
+						break;
+					}
+
+					$prefix = \str_replace( '\\\\', '[\\\\]+', \preg_quote( $prefix, '`' ) );
+					if ( \preg_match( '`^' . $prefix . '`', $content, $matches ) === 1 ) {
+						$found_prefix = $matches[0];
 						break;
 					}
 				}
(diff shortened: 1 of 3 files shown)


jrfnl:
FYI: if the hook name is single quoted and the backslashes are escaped, this is fine and won't cause a problem. Also see: https://3v4l.org/KAlvs

moorscode:
A situation I can think of that would break this is a constructed form which uses a combination of single and double quoted strings.

```php
add_action( 'Yoast\WP\' . $plugin_name . "\the_hook", ...args );
```

Not sure if the code deals with this, I don't see any tests that cover this situation.

_I would really not recommend that this kind of mixing strings is accepted, don't think that should not be the concern of this sniff._

jrfnl:
@moorscode 

> A situation I can think of that would break this is a constructed form which uses a combination of single and double quoted strings.

The backslashes are only allowed in the prefix and the prefix is supposed to be hard-coded as otherwise the hook name will be obscured for users.

Having said that, all text string parts of a hook name are examined by the sniff and there are a couple of tests covering concatenated hook names in the test file, like on line 12, 15, 42, 45, 73.

> `add_action( 'Yoast\WP\' . $plugin_name . "\the_hook", ...args );`

This snippet would not trigger the sniff as it uses `add_action()`.

Naming convention sniffs only trigger on "declarative" statements, so a class name sniff will only trigger on a class declaration, not on the _use of the class_ as that is not the part of the code where the name can be changed (and the naming of _used_ classes may be outside of our control).

So with that same reasoning, this sniff will only trigger on `do_action*()` and `apply_filters*()`, not on `add_action()`.

If I change the snippet to a declarative one and remove the parse error:

> `do_action( 'Yoast\WP\\' . $plugin_name . "\the_hook", ...args );`

... it would still not trigger as the Yoast sniff has an early check in place against the prefixes to make sure the sniff only runs on _our_ hooks and doesn't throw messages (noise) about hooks outside our control, like when a WooCommerce hook would be invoked by a Yoast plugin and the WooCommerce hook doesn't comply with the WP naming conventions for hook names.
As the prefix `Yoast\WP\` is not a valid prefix (at least not in this test case file), the sniff would bow out early for this code snippet.

Now, if I change the snippet further to actually use a valid prefix:

> `do_action( 'Yoast\WP\Plugin\\' . $name . "\the_hook", ...args );`

... now the sniff will get triggered and it would throw the following two notices:
```
 5 | WARNING | Words in hook names should be separated using underscores. Expected:
   |         | 'Yoast\WP\Plugin\\' . $plugin_name . '_the_hook', but found: 'Yoast\WP\Plugin\\' . $plugin_name . "\the_hook".
   |         | (Yoast.NamingConventions.ValidHookName.UseUnderscores)
 5 | WARNING | Hook name could not reliably be examined for maximum word count. Please verify this
   |         | hook name manually. Found: 'Yoast\WP\Plugin\\' . $plugin_name . "\the_hook"
   |         | (Yoast.NamingConventions.ValidHookName.NonString)
```

As you can see, the `"\the_hook"` code will trigger the `UseUnderscores` message as - aside from within the prefix - hooknames are not allowed to use backslashes.

> _I would really not recommend that this kind of mixing strings is accepted, don't think that should not be the concern of this sniff._

And yes, I agree, mixing string types like that is not the concern of this sniff.

If unnecessarily double quoted texts are encountered, the `Squiz.Strings.DoubleQuoteUsage` sniff would be triggered, though it wouldn't in this case due to the backslash in `"\the_hook"`, which that sniff correctly recognizes as a potential character escape which would be broken if changed to single quotes.

Does that answer your questions @moorscode ?

moorscode:
Yes, thank you! And I'm sorry for posing a question without taking some more time to look into the context/environment more 😞

Every time there is so much more going on/complexity than I realize.

jrfnl:
@moorscode No worries. I'd rather have questions which turn out to be covered already than that we'd miss something. The annoyance factor when a sniff doesn't work well enough, well you know what I mean.

moorscode (approved):

marijnyoast:
@igorschoester and I tried to acceptance test this. As there were no test instructions we tried the following:
we pulled the YoastCS repo and replaced the sniffs in our Wordpress SEO’s  vendor/yoastcs/Yoast folder with the sniffs from the branch in the PR. We then placed the three filters from [the issue](https://github.com/Yoast/yoastcs/issues/242) in a function within the WP SEO code. Afterwards we ran `vendor/bin/phpcs --standard=Yoast --sniffs=Yoast.NamingConventions.ValidHookName` {path to our file} this returned no errors while we expected one error.

jrfnl:
@marijnyoast Thanks for the attempt ;-)

> We then placed the three filters from the issue in a function within the WP SEO code. Afterwards we ran vendor/bin/phpcs --standard=Yoast --sniffs=Yoast.NamingConventions.ValidHookName {path to our file} this returned no errors while we expected one error.

Did you adjust the hook name prefixes in the example code with the prefix for YoastSEO before running the sniff ?

If not, that's the reason. See my response above to @moorscode which actually explains this as well: https://github.com/Yoast/yoastcs/pull/243#issuecomment-850287035

marijnyoast:
That could well be the cause. Hopefully the next one doing an attempt to test this will have success! 💪🏻

diedexx (approved):

PR #245, opened 2021-05-28
Add a custom "Threshold" report for use with PHPCS

jrfnl:
### Context

The addition of this report will make checking based on threshold available to all repos using YoastCS and will allow for removing some duplicate code from the YoastSEO and PremiumSEO `Composer/Actions` classes.

### Commit details

This commit adds a custom report for use with PHPCS to compare the run results with "threshold" settings.

The report will look in the runtime environment for the following two environment variables:
* `YOASTCS_THRESHOLD_ERRORS`
* `YOASTCS_THRESHOLD_WARNINGS`

... and will take the values of those as the thresholds to compare the PHPCS run results against.

If the environment variables are not set, they will default to `0` for both, i.e. no errors or warnings allowed.

To use this report, run PHPCS with the following command-line argument: `--report=YoastCS\Yoast\Reports\Threshold`.

After the report has run, a global `YOASTCS_ABOVE_THRESHOLD` constant (boolean) will be available which can be used in calling scripts, to, for instance, exit with a different exit code if the thresholds were not matched.

The report will respect the PHPCS native `colors` setting.
If `colors` is turned on, the information about a mismatched _errors_ threshold will be presented in red and for _warnings_ in yellow/orange.

Output example:
image

Note: the messaging in the report - _"update the threshold in the composer.json"_ - is strongly based on existing practices in the Yoast environment.

Includes minor tweaks to the CS rules applied to the YoastCS repo itself as this code is not run within the context of WordPress and therefore doesn't have to (and often can't) comply with WordPress specific rules.

### Testing this PR

The easiest way to test this PR is by introducing some CS errors in the code in this repo and then running the following command:
```bash
vendor/bin/phpcs --report=YoastCS\Yoast\Reports\Threshold
```

After resetting the introduced errors, run the same command again to see the "clean" report.

### DIFF
--- a/Yoast/Reports/Threshold.php
+++ b/Yoast/Reports/Threshold.php
@@ -23,6 +23,41 @@ use PHP_CodeSniffer\Reports\Report;
 class Threshold implements Report {
 
 	/**
+	 * Escape sequence for making text white on the command-line.
+	 *
+	 * @var string
+	 */
+	const WHITE = "\033[1m";
+
+	/**
+	 * Escape sequence for making text red on the command-line.
+	 *
+	 * @var string
+	 */
+	const RED = "\033[31m";
+
+	/**
+	 * Escape sequence for making text green on the command-line.
+	 *
+	 * @var string
+	 */
+	const GREEN = "\033[32m";
+
+	/**
+	 * Escape sequence for making text orange/yellow on the command-line.
+	 *
+	 * @var string
+	 */
+	const YELLOW = "\033[33m";
+
+	/**
+	 * Escape sequence for resetting the text colour.
+	 *
+	 * @var string
+	 */
+	const RESET = "\033[0m";
+
+	/**
 	 * Generate a partial report for a single processed file.
 	 *
 	 * Function should return TRUE if it printed or stored data about the file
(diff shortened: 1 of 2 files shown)


Yoast/Reports/Threshold.php
```diff
@@ -0,0 +1,125 @@
...
+			echo "Please update the WARNINGS threshold in the composer.json file to \033[32m$totalWarnings\033[0m." . \PHP_EOL;
+		}
+
+		if ( $above_threshold === false ) {
+			echo \PHP_EOL;
+			echo 'Coding standards checks have passed!' . \PHP_EOL;
+		}
+
+		// Make the threshold comparison outcome available to the calling script.
+		\define( 'YOASTCS_ABOVE_THRESHOLD', $above_threshold );
```
moorscode, line 158: Great addition!

moorscode:
Today I learned about the setup of how reports can be created and called. That's just so easy :D

Yoast/Reports/Threshold.php
```diff
@@ -0,0 +1,125 @@
...
+		$totalFixable,
+		$showSources = false,
+		$width = 80,
+		$interactive = false,
+		$toScreen = true
+	) {
+		$error_threshold   = (int) \getenv( 'YOASTCS_THRESHOLD_ERRORS' );
+		$warning_threshold = (int) \getenv( 'YOASTCS_THRESHOLD_WARNINGS' );
+
+		echo \PHP_EOL . "\033[1m" . 'PHP CODE SNIFFER THRESHOLD COMPARISON' . "\033[0m" . \PHP_EOL;
```
moorscode: I'd prefer to have the colors as variables (red, green, orange, reset?), to avoid additional comments and have a better "sight" of what the expect output will be.

moorscode (changes requested):
Looks clean and simple, just one request for a bit more explicit use of the colors in the output.

Yoast/Reports/Threshold.php
jrfnl: Thanks. This is atypical for a PHPCS report, but seemed a reasonable thing to do for our purposes. I have commits ready to implement the use of the reports (and this constant) for the Free and Premium plugins (and to remove parts of the code now in the composer/actions files).

Yoast/Reports/Threshold.php
jrfnl: @moorscode I hear you.

The reason I didn't do this is that there will be a function for colourizing text (with the colour name as one of the parameters) in the near future in PHPCSUtils and once we switch to using PHPCSUtils, we can use that function.

What about if I - for now - set up the colour as class constants - self::RED - and refer to them that way ? That would make them descriptive for the short term.

moorscode: Sounds great!

jrfnl: Done (in a separate commit for easier review).

moorscode:
Looks good to me!

_I spot some additional tree saving in the last commit as well, dot to comma_ 😁

jrfnl:
@moorscode Yup ;-) as per the commit message:

>  Threshold report: use colour constants
> 
> This changes the colourization of the text messages to be more descriptive by using class constants for the colour codes.
>
> As a lot of the `echo` statements can now no longer use double quoted strings (due to the references to class constants), this also changes the `echo` statements to use comma separated arguments instead of concatenation for creating the output.

moorscode (approved):

JoseeWouters:
When trying to run the command specified in the testinstructions, I get an error: `ERROR: Class file for report "YoastCSYoastReportsThreshold" not found`. I've tried some variations, but they all result in `Class file for report not found`. Am I missing something?

jrfnl:
> Am I missing something?

@JoseeWouters Possibly - did you run `composer install` after checking out this repo ? Or did you try to test with another repo ?

jrfnl:
Oh hang on - what OS are you on ? You may need to escape the backslashes, like so:
```bash
# Put the value within quotes:
vendor/bin/phpcs --report="YoastCS\Yoast\Reports\Threshold"
# Or escape the backslashes: 
vendor/bin/phpcs --report=YoastCS\\Yoast\\Reports\\Threshold
```

JoseeWouters:
I'm on MacOS. Had run a composer install and Commands in the readme do work for me. Will try this!

JoseeWouters:
Yes, escaping worked. 
Acceptance: 👍

PR #251, opened 2021-07-28
Commenting/CodeCoverageIgnoreDeprecated: prevent false negatives with PHP 8.0 attributes

jrfnl:
PHP 8.0+ attributes can be placed between a docblock and the function declaration it applies to.

The `Yoast.Commenting.CodeCoverageIgnoreDeprecated` sniff did not yet take this into account.

This would result in a false negatives as the sniff would bow out for not finding a docblock, even when the docblock contained a `@deprecated` tag.

Fixed now.

Includes unit tests.

### DIFF
--- a/Yoast/Sniffs/Commenting/CodeCoverageIgnoreDeprecatedSniff.php
+++ b/Yoast/Sniffs/Commenting/CodeCoverageIgnoreDeprecatedSniff.php
@@ -39,17 +39,27 @@ class CodeCoverageIgnoreDeprecatedSniff implements Sniff {
 	public function process( File $phpcsFile, $stackPtr ) {
 
 		$tokens = $phpcsFile->getTokens();
-		$find   = Tokens::$methodPrefixes;
-		$find[] = \T_WHITESPACE;
 
-		$commentEnd = $stackPtr;
-		do {
-			$commentEnd = $phpcsFile->findPrevious( $find, ( $commentEnd - 1 ), null, true );
-		} while ( $tokens[ $commentEnd ]['line'] === $tokens[ $stackPtr ]['line'] );
+		$ignore                  = Tokens::$methodPrefixes;
+		$ignore[ \T_WHITESPACE ] = \T_WHITESPACE;
 
-		if ( $tokens[ $commentEnd ]['code'] !== \T_DOC_COMMENT_CLOSE_TAG
-			|| $tokens[ $commentEnd ]['line'] !== ( $tokens[ $stackPtr ]['line'] - 1 )
-		) {
+		for ( $commentEnd = ( $stackPtr - 1 ); $commentEnd >= 0; $commentEnd-- ) {
+			if ( isset( $ignore[ $tokens[ $commentEnd ]['code'] ] ) === true ) {
+				continue;
+			}
+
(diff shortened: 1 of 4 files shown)


increddibelly:
Hi @jrfnl, we've looked at the PR and we managed to decipher it :sweat_smile:
we thought of an edge case that you might want to consider in additional tests:

```PHP
/**
 * docblock
 */
// I added a fietsbel decorator                              <- this is valid PHP
#[fietsbel] // added a fietsbel decorator                    <- so is this
function TestFunction( $henk ) {
...
}
```
even though it isn't pretty, should the sniff be able to deal with it?

increddibelly (approved):

jrfnl:
Hi @increddibelly 👋🏻 

Thanks for reviewing this PR!


> we thought of an edge case that you might want to consider in additional tests:
> 
> ```
> /**
>  * docblock
>  */
> // I added a fietsbel decorator                              <- this is valid PHP
> #[fietsbel] // added a fietsbel decorator                    <- so is this
> function TestFunction( $henk ) {
> ...
> }
> ```
> 
> even though it isn't pretty, should the sniff be able to deal with it?


As a rule of thumb there should be *nothing* between a docblock and the function declaration it applies to. 

The sniff previously allowed for blank lines (another sniff is in place to complain about this if needs be) and this PR adds an allowance for PHP 8.0 attributes (also added to the other sniff(s)).

While I understand that you may want to add an inline comment to an attribute, when the attribute applies to a function declaration, as things are, this is not allowed, as the sniff will not be able to tell based on the code whether the comment is an incorrectly formatted docblock for the function declaration or whether the comment applies to the attribute.

Luckily there is a solution if you do want to use inline comments for attributes which apply to function declarations: turn the order around:

```php
// I added a fietsbel decorator
#[fietsbel] // added a fietsbel decorator
/**
 * docblock
 */
function TestFunction( $henk ) {}
```

PHP strips docblocks and comments when compiling, so whether the attribute is before or after the docblock doesn't actually matter for PHP.

Also know that comments can also be used _within_ attributes and in that case, it will be ignored by the code now added, though the below multi-line format would cause a parse error in PHP < 8.0.
```php
/**
 * docblock
 */
#[
    // I added a fietsbel decorator
    fietsbel  // added a fietsbel decorator
]
function TestFunction( $henk ) {}
```

This second example would be okay cross-version though:
```php
/**
 * docblock
 */
#[fietsbel  /* added a fietsbel decorator */]
function TestFunction( $henk ) {}
```


So having said that, I don't think allowing for inline comments for attributes should currently be added.

I will, however, keep your comment in mind for PHPCSUtils, which will contain utility functions to find a docblock and I think allowing trailing comments to an attribute should possibly be supported in those utilities.

Does that sound like a good enough compromise to you ?

increddibelly:
thank you for that great explanation! IMHO the only thing we could improve on this PR, is to add a testcase that shows how this sniff responds to the bad&ugly inline docblock comments, simply to document the current behavior, even if it's desired behavior or not, since it's such an edge case.

jrfnl:
@increddibelly I've updated the PR to contain the extra test cases documenting the behaviour in combination with inline comments related to attributes. Should be good to go now.

PR #252, opened 2021-07-28
Commenting/CoversTag: make more code-style independent

jrfnl:
The regex used in the `Yoast.Commenting.CoversTag` sniff was - in part - based on the known naming conventions used in WP projects.

As the Yoast standard will now also be used outside of WP projects (special projects team), the assumption that the WP naming conventions will be used needs to be removed.

Fixed now.

Includes updated and additional unit tests which would not pass without this fix.

### DIFF
--- a/Yoast/Sniffs/Commenting/CoversTagSniff.php
+++ b/Yoast/Sniffs/Commenting/CoversTagSniff.php
@@ -24,11 +24,9 @@ class CoversTagSniff implements Sniff {
 	/**
 	 * Regex to check for valid content of a @covers tags.
 	 *
-	 * Takes the WP naming conventions into account (up to a point).
-	 *
 	 * @var string
 	 */
-	const VALID_CONTENT_REGEX = '(?:\\\\?(?:(?<OOName>[A-Z][a-zA-Z0-9_\x7f-\xff]*)\\\\)*(?P>OOName)(?:<extended>|::<[!]?(?:public|protected|private)>|::(?<functionName>(?!public$|protected$|private$)[a-z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*))?|::(?P>functionName)|\\\\?(?:(?P>OOName)\\\\)+(?P>functionName))';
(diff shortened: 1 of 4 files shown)


Yoast/Sniffs/Commenting/CoversTagSniff.php
```diff
@@ -24,11 +24,9 @@ class CoversTagSniff implements Sniff {
 	/**
 	 * Regex to check for valid content of a @covers tags.
 	 *
-	 * Takes the WP naming conventions into account (up to a point).
-	 *
 	 * @var string
 	 */
-	const VALID_CONTENT_REGEX = '(?:\\\\?(?:(?<OOName>[A-Z][a-zA-Z0-9_\x7f-\xff]*)\\\\)*(?P>OOName)(?:<extended>|::<[!]?(?:public|protected|private)>|::(?<functionName>(?!public$|protected$|private$)[a-z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*))?|::(?P>functionName)|\\\\?(?:(?P>OOName)\\\\)+(?P>functionName))';
+	const VALID_CONTENT_REGEX = '(?:\\\\?(?:(?<OOName>[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*)\\\\)*(?P>OOName)(?:<extended>|::<[!]?(?:public|protected|private)>|::(?<functionName>(?!public$|protected$|private$)(?P>OOName)))?|::(?P>functionName)|\\\\?(?:(?P>OOName)\\\\)+(?P>functionName))';
```
moorscode, line 29: Could you explain what this does:

`[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*` versus `[A-Z][a-zA-Z0-9_\x7f-\xff]*`

The internet tells me it's extending the range of allowed characters beyond ASCII; is this for multibyte language support or am I thinking too far out?

Or is this just to make sure it doesn't have to start with a capital letter? 😮

jrfnl: This is the official regex as per the PHP documentation for what is allowed as an object/function name.
It will now allow for all PHP supported names, not just the ones allowed according to the WP coding standards.
 
Example which is perfectly valid PHP (but please don't use):
```php
class 💩💩💩💩
{
	function 💩💩💩($😎, $🐯)
	{
		return $😎 + $🐯;
	}

}
$🐔 = 3;
$😥 = $🐔 + 2;
$💩 = new 💩💩💩💩;
echo $💩 -> 💩💩💩($🐔, $😥);
```

Refs;
* https://www.php.net/manual/en/language.oop5.basic.php
* https://www.php.net/manual/en/functions.user-defined.php

moorscode: Thanks, that is just so simple, and also very terrifying (the example).

moorscode (approved):
Code and tests check out 👍

PR #253, opened 2021-07-28
Commenting/FileComment: efficiency tweak + bug fix

jrfnl:
### Commenting/FileComment: efficiency tweak

While reviewing the sniffs for WP-code style independence (for use in special projects), I noticed that the `FileComment` sniff may walk all the way to the end of a file before deciding to pass off to the parent sniff.

While for namespaced files, this will never happen as a namespace declaration will (normally) be the first statement in a file, for non-namespaced files, this made the sniff highly inefficient as, especially for long files, it would have to walk through all the tokens in the file before deciding it is a non-namespaced file.

Fixed now by only walking as far as the first non-empty (not whitespace, not comment) token and deciding based on that.

While a file _may_ contain a namespace declaration later on in the file (multi-namespace files), this rule is not intended for those type of files anyway, so it is correct for the "must-have file comment" rule to be applied in that case.

This change is already covered by the existing unit tests.

### Commenting/FileComment: minor bug fix

The PHPCS `File::findNext()` function will return the stack pointer to the token found or `false` if no matching token is found.

The check for the `$comment_start` token could therefore return `false` when no comment token is found between a PHP open tag and the first namespace token, which would mean that the `$tokens[ $comment_start ]['code']` check in the condition would effectively look at `$tokens[ false ]['code']`, which (in most cases) would translate to the PHP open tag.

While this doesn't lead to any problems in practice, it is still wrong, so a check against `false` has been added to the condition now.

### DIFF
--- a/Yoast/Sniffs/Commenting/FileCommentSniff.php
+++ b/Yoast/Sniffs/Commenting/FileCommentSniff.php
@@ -82,7 +82,7 @@ class FileCommentSniff extends Squiz_FileCommentSniff {
 
 		$comment_start = $phpcsFile->findNext( \T_WHITESPACE, ( $stackPtr + 1 ), $namespace_token, true );
 
-		if ( $tokens[ $comment_start ]['code'] === \T_DOC_COMMENT_OPEN_TAG ) {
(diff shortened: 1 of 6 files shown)


moorscode (approved):
Code looks good and tests seem to cover all bases.

PR #257, opened 2021-08-17
Allow Exception namespace

joopyoast:
Allow an extra Namespace so we can have an `App/Context/Object/Exception/` structure

I also added .idea/ to the ignore files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.lock
 .cache/phpcs.cache
 phpcs.xml
 phpunit.xml
+.idea/
--- a/Yoast/Sniffs/NamingConventions/NamespaceNameSniff.php
+++ b/Yoast/Sniffs/NamingConventions/NamespaceNameSniff.php
@@ -50,7 +50,7 @@ class NamespaceNameSniff implements Sniff {
 	 *
 	 * @var int
 	 */
-	public $max_levels = 3;
+	public $max_levels = 4;
 
 	/**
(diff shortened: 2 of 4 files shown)


jrfnl:
@joopyoast Sorry, but this PR does not belong here.

The `recommended_max_levels` and `max_levels` properties are in place to allow for the **_actual_** allowed level to be adjusted in project rulesets.
So not sure what project you are currently working on, but you can adjust the property in the PHPCS ruleset of the project like so:
```xml
	<rule ref="Yoast.NamingConventions.NamespaceName">
		<properties>
			<property name="recommended_max_levels" value="3"/>
			<property name="max_levels" value="4"/>
		</properties>
	</rule>
```

Aside from that:
Typically an ignore for an IDE specific folder, like `.idea`, belongs in a dev's `.git/info/excludes`. This project does not specify the IDE which "must" be used, so shouldn't account for it in these kind of files either.

joopyoast:
Hi Juliette,

Thanks for taking time to point me in the correct direction.
I am working on one of the projects from Yoast so I thought this would be the correct place.

jrfnl:
@joopyoast No worries, I understand this can be confusing, but this ruleset is used for all Yoast projects so a change affects them all, not just the project you are working on.
If you ever have more questions about YoastCS, know that I'm also in the Yoast Slack and there's even a dedicated #yoastcs channel. Please feel free to ping me and to ask any questions you have there and I'll try to help you.

PR #273, opened 2022-02-11
New `BrainMonkeyRaceCondition` sniff

### LINKED ISSUE #264: Detect potential race condition with BrainMonkey test functions
As discussed with @increddibelly:

1. The potential race condition needs to be reported to BrainMonkey and if possible, fixed there.
2. In the mean time, we want to prevent the Yoast tests running into it.

### Issue outline

> Functions\expect( 'apply_filters' )  should not be used; it causes a race condition with Filters\expectApplied( 'filter_name' ).
> A relatively simple sniff could prevent that.
> The same goes for actions: with expect( 'do_action' ) (bad) and expectDone (good)

### Sniff outline

* Register `T_STRING` and check for content `expect`
* with a `T_NS_SEPARATOR` before it and `T_STRING` `Functions` before that and `T_OPEN_PARENTHESIS` after
* first arg of function call should be `T_CONSTANT_ENCAPSED_STRING` with content `apply_filters` or `do_action`
* check last relevant (function, closure, oo) condition and last condition is a function/closure.
* MAYBE -> is this a function in a test class ? (doesn't need to be a test function, can also be a test helper function) -> _eigenlijk niet relevant -> check bij schrijven sniff of het een verschil maakt_
* if so, check if `expectApplied` of `expectDone` is called within the same function
    in other words: `T_STRING` met content `expect...` + followed by open parenthesis

### Planning

=> YoastCS 2.3.0 + tag @increddibelly as reviewer + inform Irene about release

jrfnl:
This new sniffs detects a particular race condition which can occur when writing tests using the BrainMonkey utilities and which causes tests to fail.

Fixes #264

Includes tests.
Includes docs.

### DIFF
--- a/Yoast/Docs/Tools/BrainMonkeyRaceConditionStandard.xml
+++ b/Yoast/Docs/Tools/BrainMonkeyRaceConditionStandard.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0"?>
+<documentation title="BrainMonkey Race Condition">
+    <standard>
+    <![CDATA[
+    When writing tests using the BrainMonkey test utilities, either mock and set an expectation for calls to WP hook functions using `Monkey\Functions\expect()` or set the expectation using `Monkey\Filters\expectApplied()` or `Monkey\Actions\expectDone()`.
+    
+    Do not mix the use of these two types of setting expectations within the same test method as that will cause the tests to fail.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: using either of the function, but not mixing them in the same test.">
+        <![CDATA[
+class MyTest extends TestCase {
+  public function test_using_only_expect() {
+    expect( 'apply_filters' )->once();
+    Functions\expect( "apply_filters" )
+      ->once();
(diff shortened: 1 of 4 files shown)


jrfnl:
Rebased without changes.

diedexx (commented):
Looking at the code, I suspect that the following combination doesn't cause the race condition. Do you think it's worth adding a test-case to document that behaviour?
``` php
Monkey\Actions\expectDone( 'action_name' )->with( $param );
Monkey\Functions\expect( 'apply_filters' )->once();
```

jrfnl:
> Looking at the code, I suspect that the following combination doesn't cause the race condition. Do you think it's worth adding a test-case to document that behaviour?
> 
> ```
> Monkey\Actions\expectDone( 'action_name' )->with( $param );
> Monkey\Functions\expect( 'apply_filters' )->once();
> ```

Very good point! And yes, the sniff already handles this correctly, but I totally agree it would be good to safeguard this for the future!

Added extra test cases via a new commit.

diedexx (approved):

PR #372, opened 2023-12-18
Sets the minimum supported WordPress version to 6.3.

thijsoo:
This PR bumps the minimum supported WordPress version to 6.3, since we are also dropping support for 6.2 in all the plugins.


jrfnl (approved):
Thanks @thijsoo !

PR #380, opened 2024-05-27
Sets the minimum supported WordPress version to 6.4.

enricobattocchi:
This PR bumps the minimum supported WordPress version to 6.4, since we are also dropping support for 6.4 in all the plugins.


jrfnl:
> This PR bumps the minimum supported WordPress version to 6.4, since we are also dropping support for 6.4 in all the plugins.

I presume you mean "dropping support for 6.3" ?

jrfnl (approved):

PR #381, opened 2024-10-01
YoastCS ruleset: set the minimum supported WP version to 6.5

leonidasmi:
This PR bumps the minimum supported WordPress version to 6.5, since we are also dropping support for 6.5 in all the plugins


jrfnl (approved):

jrfnl:
Rebased without changes to get a passing build.

